PR #10, opened 2021-04-12
refactor blockpage detection and maxminddb to work in the cloud

ohnorobo:
Some refactoring and various changes to get the pipeline into a state where it can be successfully deployed to prod for the existing 4 scan types.
This is an intermediate step to getting the full blockpage and dns parts of the pipeline working.

Deleted:
- censys code, this wasn't being used, we can add it back later.
- the signature files, just kept some small sample files for tests. The full files are read in from the cloud.

Refactored:
- Maxmind database into its own object which reads data from the cloud.
- Blockpage detection into its own object which reads data from the cloud.
- country_name -> country_code into a little side function with data inlined.

Added
- e2e tests for various things, and kept existing unit tests even for pieces I turned off.
- maxmind files in GCS
- signature files in GCS

Turned off
- dns runs as part of the `--all` pipeline runs. This doesn't yet work in the cloud.
- passing the actual blockpage matcher to pipeline runs. (I'm just passing a null side arg instead). This doesn't yet work in the cloud, and will have to be reconfigured per https://github.com/censoredplanet/jigsaw-cp-tracker/issues/1#issuecomment-818047521

manual testing: ran http/s dev pipelines and saw them succeed.


ohnorobo:
@avirkud we can chat about the changes here when we meet tomorrow

pipeline/test_beam_tables.py
```diff
@@ -728,16 +756,16 @@ def test_merge_metadata_with_rows(self) -> None:
...
ç¾è³è¨å®å
¨é²è­· - æ¡æç¶²é é»æ\u003c/h1\u003e\r\n\u003cp\u003eå¦ææ¨çå°è©²ç«é¢è³è¨ï¼è¡¨ç¤ºæ¨è¢«å¤æ·å­åéæ­£å¸¸è¡çºç¶²ç« \u003cspan style=\"color:red;\"\u003e(æ¡æç¶²ç«)\u003c/span\u003e\u003c/p\u003e\r\n\u003cp\u003eè³è¨èª²å·²å°æ­¤ç¶²é é»æï¼å¦æç¢ºå®è©²ç¶²é æ¯è¢«èª¤å¤è«è¯ç¹«: é
ç¾è³è¨èª²-ç³»çµ±ç¶­éçµï¼è¬è¬ã\u003c/p\u003e\r\n\u003cp\u003e\u003cb\u003eä½¿ç¨è
:\u003c/b\u003e 141.212.123.175 \u003c/p\u003e\r\n\u003cp\u003e\u003cb\u003eç¶²å:\u003c/b\u003e rtyutgyhefdafioasfjhjhi.com/ \u003c/p\u003e\r\n\u003cp\u003e\u003cb\u003eåé¡:\u003c/b\u003e command-and-control \u003c/p\u003e\r\n\u003c/div\u003e\r\n\u003c/body\u003e\r\n\u003c/html\u003e\r\n'''
```
ohnorobo, line 740: This is one area where I feel like I've made the code worse. I changed some test cases because I'm only including the reduced signature files and so can only pick up a few patterns. (I tried to find the matching signature lines for the original pages, but in some cases I wasn't able to.)

I think your old test cases here are better and exercise more interesting things. If you can help me figure out which exact signature lines match your older test cases here I'd like to switch back to them.

avirkud: These are the patterns from the original test cases:
1. (false positive) `{"fingerprint":"x_generic_akamai_error","pattern":"An error occurred while processing your request.<p>"}`
2. (blockpage) `{"fingerprint":"a_prod_drweb_3","pattern":"URL % Sp%er Gate"}`
3. (no match)
4. (blockpage) `{"fingerprint":"c_isp_ru_iterika","pattern":"iterika.ru/blocked.html?UrlRedir=http%3A%2F%2Fblackberry.com%2f"}`
5. (blockpage) `{"fingerprint":"c_isp_ru_divo_2","pattern":"https://www.divo.ru/"}`
6. (blockpage) `{"fingerprint":"e_unk_style_red","pattern":"<span style=%color:red;%>(æ¡æ"}`

ohnorobo: Thanks, I've added these and switched back to the old test cases.

avirkud (approved):

PR #12, opened 2021-04-16
package blockpage data with module and match in flatten step

ohnorobo:
This change
- packages the blockpage files with the module, initializes the blockpage class with those files in-place instead of reading from the cloud.
- converts the flatten step into a `DoFn` class
- initializes the blockpage matcher in the `setup` stage of the DoFn, so it will only be run once per worker
- use updates blockpage files with per-escaped regular expressions


pipeline/metadata/blockpage.py
```diff
@@ -21,7 +21,7 @@ def _load_signatures(filepath: str) -> Dict[str, re.Pattern]:
   """
   data = pkgutil.get_data(__name__, filepath)
   if not data:
-    raise FileNotFoundError(f"Couldn't find file {filename}")
+    raise FileNotFoundError(f"Couldn't find file {filepath}")
 
   content = data.decode('utf-8')
   lines = content.split('\n')
```
fortuna, line 27: You have 3 copies of the file: `data`, `content` and `lines`.
Instead, wrap data in StringIO and read each line

pipeline/beam_tables.py
```diff
@@ -310,76 +310,103 @@ def _parse_received_headers(headers: Dict[str, List[str]]) -> List[str]:
...
+    Args:
+      filename: a filepath string
+      scan: a loaded json object containing the parsed content of the line
+      random_measurement_id: a hex id identifying this individual measurement
+
+    Yields:
+      Rows
+    """
     date = re.findall(r'\d\d\d\d-\d\d-\d\d', filename)[0]
     if date < "2021-03":
```
fortuna, line 411: This kind of thing can make the code hard to follow. I'd have 3 methods:
- formatVersion(filename) or isV1(filename)
- parseV1Line(line)
- parsev2Line(line)

pipeline/metadata/blockpage.py
```diff
@@ -1,55 +1,57 @@
...
 
 # Signature filenames
-FALSE_POSITIVES = 'false_positive_signatures.json'
-BLOCKPAGES = 'blockpage_signatures.json'
+FALSE_POSITIVES = 'data/false_positive_signatures.json'
+BLOCKPAGES = 'data/blockpage_signatures.json'
+
+
+def _create_regex(pattern: str) -> Pattern:
+  pattern = re.escape(pattern)
```
fortuna, line 14: You need to \\-unescape the string, as specified here: https://cloud.google.com/bigquery/docs/reference/standard-sql/operators#comparison_operators

But the original `%`s should be replaced with `.*` and `_` with `.`

My guess is that the original patterns are not properly \\-escaped, so you need to validate and fix the patterns too.

fortuna (changes requested):

pipeline/metadata/blockpage.py
ohnorobo: done

pipeline/beam_tables.py
ohnorobo: done

pipeline/metadata/blockpage.py
ohnorobo: No, this ended up not being the problem. The re.escape line is already escaping most things (like `.` and spaces) correctly.

`%` is being turned correctly into `.*` in the final regex
`_` should stay as `_`. If it was operating as a special char in bq I think that was unintended. (if you look at the patterns it's clearly not intended as a regex character)

I'll make a longer comment below.

ohnorobo:
Alright. I've got this working. ([https run](https://pantheon.corp.google.com/dataflow/jobs/us-east1/2021-04-21_13_14_27-4272829860443673088;step=s10?project=firehook-censoredplanet), [http run](https://pantheon.corp.google.com/dataflow/jobs/us-east4/2021-04-21_15_34_46-9691489344334171244;step=s10?project=firehook-censoredplanet))

The problem was a pathological interaction between regexes of the form `.*sometext.*` and using the `re.search` option. Even though technically that should just be redundant in practice it causes the regex to blow up. There were only a few cases that were doing this and I've removed them in https://github.com/censoredplanet/censoredplanet-analysis/pull/12/commits/2dff1d3ef634f25f9977707bc809fc2a23ce3b92. I also cleaned up a few other cases of `%` where it was just being used to avoid special chars like `'`. @ramakrishnansr would you mind taking a look at the changes I've made to the signature files in [that commit](https://github.com/censoredplanet/censoredplanet-analysis/pull/12/commits/2dff1d3ef634f25f9977707bc809fc2a23ce3b92) and letting me know if they look alright to you?

pipeline/metadata/blockpage.py
ohnorobo: The url-escaped `%` chars also ended up being a red herring, although they're still technically being escaped incorrectly. As long as the regex has some unique string prefix like `https://www.divo.ru/` it starts by looking for that. Since it never finds it it never gets to the expensive many-`.*` portion of the regex.

fortuna: The code and data are still broken. We need to fix both.

First, we need to precisely define what the input format is. It's currently undefined. Is it a BigQuery LIKE string? Is it a regex?
It's currently none, since it's not a regex and we are not properly escaping as LIKE, as there are multiple `_` and `%` with the wrong semantics. We must escape them if we want to declare the input is the LIKE format.
`UrlRedir=http%3A%2F%2F1337x.to%2f` should be `UrlRedir=http\\%3A\\%2F\\%2F1337x.to\\%2f` in the json file (`UrlRedir=http\%3A\%2F\%2F1337x.to\%2f` as the parsed string).

My suggestion is to replace the pattern string with a list of strings. This way you can easily generate a LIKE string by replacing all `_\%` with the escaped version for all parts and join the result with `%`.
And you can easily generate a regex by re-escaping the parts and joining with `.*`.
Otherwise you will need to do the split, unescape, replace, re-escape, join.

Another possibility is for us to come up with our own format, where we have a special, unique mark as the place to insert the `.*` or `%`. For example, insert a mark like `$BREAK$`, assuming it doesn't occur in the patterns. Then you escape the whole pattern, and replace the escaped `$BREAK$` (`\$BREAK\$` in RE, `$BREAK$` in LIKE) with the appropriate operator.

In short, the 3 options are:

1. Use a list of strings in the patterns
2. Use LIKE format and do the split-unescape-escape-join dance
3. Use the original strings, but with a unique marker you can replace (not `%`)

@ramakrishnansr this is something that must be fixed in the pattern generation too, since it's generating patterns that are misusing the special characters.

fortuna (changes requested):

pipeline/metadata/blockpage.py
fortuna: There's a fourth option: save the file in regex format. Then you just read and use.
If we end up needing to use the patterns in BigQuery, we can call REGEXP_CONTAINS instead of LIKE.

ramakrishnansr: I think the last option makes the most sense. I can spend some time today making sure all the patterns are in regex format and push that to our assets. Then all of the problems can be avoided.

ramakrishnansr (changes requested):
I think there's also potential to add [The Citizen Lab's blockpage fingerprints](https://github.com/citizenlab/filtering-annotations/blob/master/analysislib/known_blockpages.py) and [OONI's blockpage fingerprints](https://github.com/ooni/pipeline/blob/master/af/fastpath/fastpath/utils.py) for extra data. I don't know if we have the bandwidth to do that currently, but worth exploring, I think. Meanwhile, I will make a revision of our patterns on our assets website.

pipeline/metadata/blockpage.py
fortuna: I agree. Anyone will be able to use and contribute with the regex format.

ohnorobo:
Updated to the newest blockpage files. Please take a look. @fortuna, @ramakrishnansr

pipeline/metadata/blockpage.py
```diff
@@ -1,55 +1,48 @@
...
-        Either a gcs filepath or a local system folder.
-    """
-    false_positive_path = os.path.join(signature_folder, FALSE_POSITIVES)
-    blockpage_path = os.path.join(signature_folder, BLOCKPAGES)
-
-    self.false_positives = _load_signatures(false_positive_path)
-    self.blockpages = _load_signatures(blockpage_path)
+  def __init__(self) -> None:
+    """Create a Blockpage Matcher."""
+    self.false_positives = _load_signatures(FALSE_POSITIVES)
```
fortuna, line 44: Because the data is hardcoded, it makes it hard to test and reuse this class.

Instead:

- make constructor create an empty matcher 
- provide `add_blockpage(id, pattern)` and `add_false_positive(id, pattern)` methods
- provide a `create_default_matcher()` that reads from the default location and calls the add methods.

I'd recommend writing a pattern generator that you can reuse. Something like:
```py
def read_patterns(data):
  for line in data:
    # parse
    yield (id, pattern)
```
So you can have:
```py
for id, pattern in read_patterns(pkgutil.get_data(__name__, filepath)):
  matcher.add_blockpage(id, pattern)
```

pipeline/beam_tables.py
```diff
@@ -448,15 +386,151 @@ def _flatten_measurement(
...
+      a dict containing the 'received_' keys/values in SCAN_BIGQUERY_SCHEMA
+    """
+    if isinstance(received, str):
+      return {'received_status': received}
+
+    row = {
+        'received_status':
+            received['status_line'],
+        'received_body':
+            received['body'],
```
fortuna, line 513: I'm surprised this doesn't crash with a KeyError.

fortuna (changes requested):

pipeline/beam_tables.py
```diff
@@ -448,15 +386,151 @@ def _flatten_measurement(
...
+        'error': scan.get('error', None),
+        'blocked': not scan['passed'] if 'passed' in scan else None,
+        'success': 'error' not in scan,
+        'received': None,
+        'measurement_id': random_measurement_id,
+    }
+    received_ips = scan.get('answers')
+    yield from self._process_received_ips(row, received_ips)
+
+  def _process_satellite_v2(self, scan: Any,
```
ramakrishnansr, line 440: @avirkud For Satellite-v2, we might want to do blockpage matching using the `blockpages.json` file.

ramakrishnansr (changes requested):

pipeline/metadata/blockpage.py
```diff
@@ -1,55 +1,48 @@
...
-    path: path to json file containing signatures
+    filepath: relative path to json file containing signatures
 
   Returns:
     Dictionary mapping fingerprints to signature patterns
   """
+  data = pkgutil.get_data(__name__, filepath)
+  if not data:
+    raise FileNotFoundError(f"Couldn't find file {filepath}")
+  content = io.TextIOWrapper(io.BytesIO(data), encoding='utf-8')
```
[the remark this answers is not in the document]
ohnorobo: No, this errors with 
`AttributeError: 'bytes' object has no attribute 'readable' [while running 'flatten json']`

pipeline/beam_tables.py
ohnorobo: Some of the fields are just very consistent and exist in 100% of the data.

pipeline/metadata/blockpage.py
ohnorobo: Strong disagree. As long as the data is inlined and inspectable I think this adds too much pattern overhead. The factory pattern is worth doing when we're reading data from an external cloud-project-level source, but it adds data-passing overhead which we already have a lot of.

I don't like the idea of having a constructor which creates an empty object which you have to do a bunch of extra setup to use. (Even if that setup can be encapsulated inside of a factory method.) It just adds complexity.

Tellingly, even if I made this change I don't think I would change any of the examples in `test_blockpage_matching` or `test_blockpage.py`. I'm already confident in the testing we're able to do for this class with the data inlined.

pipeline/metadata/blockpage.py
```diff
@@ -1,55 +1,48 @@
...
-    blockpage_path = os.path.join(signature_folder, BLOCKPAGES)
-
-    self.false_positives = _load_signatures(false_positive_path)
-    self.blockpages = _load_signatures(blockpage_path)
+  def __init__(self) -> None:
+    """Create a Blockpage Matcher."""
+    self.false_positives = _load_signatures(FALSE_POSITIVES)
+    self.blockpages = _load_signatures(BLOCKPAGES)
 
   def match_page(self, page: str) -> Optional[bool]:
```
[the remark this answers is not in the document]
ohnorobo: My thought was to use the existing interface for this PR and then update it after since there's already a lot changing here.

What are your thoughts on the right data to expose to users? My initial thought was to have only a `blockpage_signature` field and not expose the FP data since end-users don't care about the details of our FP detection. But we're using this internally too, and *we* care, since we want to be able to debug the system.

Maybe have the columns be
`blockpage`: `<pattern string>` which would be for end users/dashboards
`false_positive_blockpage`: `<pattern string>` which would be for internal debugging.

pipeline/beam_tables.py
avirkud: Yes, I can add this once the current Satellite pipeline is working in the cloud.

pipeline/metadata/blockpage.py
fortuna: What complexity are you referring to?
It doesn't add complexity to the client at all. Instead of calling BlockpageMatcher() you call create_default_blockpage_matcher().
Meanwhile, the tests are hard to read, since you need to know what patterns are being tested. I can't tell what's being tested without searching some other file. Furthermore, there's an unnecessary dependency on pkgutil and a hidden dependency on a global (the filepath) too deep in the code.

I just realized that this class has another problem: It does too much. It violates the single-responsibility principle.
Instead, you should have a simple PageMatcher() that all it does is take a page and return the pattern id. There's no reason for it to know about blockpages and false positivies. Then in the processing code you can have a matcher for blockpages and one for false positives. Rename `_load_signatures` to `create_matcher_from_file`.
This makes the concepts simpler and easier to understand, besides removing duplicate code and making the code more reusable and the tests more self-contained.

Because of the size of this PR, I'd be ok to fix this in a follow-up PR. But I'd like to see that change. We shouldn't mix low level and high level concepts. It makes understanding harder.

pipeline/metadata/blockpage.py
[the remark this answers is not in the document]
fortuna: I agree we should expose the signature id. Let's keep it as a TODO for now.

I would suggest `page_signature: <pattern_id>, is_block_page:<true|false|unknown>`
`page_signature` is the id, not the pattern.`is_block_page` can be derived from `page_signature`.

This will allow us to generate explanations like `HTTP Interference: domain $domain returns blockpage $page_signature` 

Also, in the future I can see use tagging pages even if they are not blockpages, like the cloudflare ddos protection page.

fortuna (approved):

pipeline/metadata/blockpage.py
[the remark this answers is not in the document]
fortuna: By the way, having a way to tag generic pages (like Cloudflare) is another reason to have the simpler PageMatcher.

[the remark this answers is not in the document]
ohnorobo: The reason I was thinking of having `blockpage` / `false_positive_blockpage` was for the sake of our future dashboard use. If both blockpages and false positive signature are mixed up in the same column it will be difficult to display them separately, or to only display blockpages.

But I agree this can all be a TODO. It will be easier to reason about in a follow-up PR.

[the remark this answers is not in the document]
fortuna: The reason I proposed the `is_block_page` is for the dashboard or a data consumer. The page fingerprint would be mostly for inspection, validation, to see if things are consistent.

pipeline/metadata/blockpage.py
ohnorobo: added a TODO

pipeline/metadata/blockpage.py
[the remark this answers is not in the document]
ohnorobo: added a TODO

PR #14, opened 2021-04-23
split out satellite bq schema fields to only apply to satellite pipelines

ohnorobo:
I was trying to re-run the quack/hyperquack pipelines from head in this repo. But I realized that the new field for the satellite schema mean that we can't write successfully to the existing tables. It fails on schema mismatch errors.

Since we're not currently running the satellite pipeline I've broken out those field into a seperate part of the schema. Eventually we may want to recombine them so all the tables are uniform, but for now this should make it easier to keep developing the satellite pipeline without having to backfill the other pipelines any time we make schema changes. 

I've also removed the `blockpage` field in the schema for now. But I'll add it back in https://github.com/censoredplanet/censoredplanet-analysis/pull/12.

Testing: ran [catch up jobs](https://console.cloud.google.com/dataflow/jobs?project=firehook-censoredplanet) for the 4 existing pipelines.

### DIFF
--- a/pipeline/metadata/maxmind.py
+++ b/pipeline/metadata/maxmind.py
@@ -28,12 +28,12 @@ def _maxmind_reader(filepath: str) -> geoip2.database.Reader:
   # MaxMind Reader will only take a filepath,
   # so we need to write the file to local disk
   disk_filename = os.path.join('/tmp', os.path.basename(filepath))
-  disk_file = open(disk_filename, 'wb')
-  disk_file.write(f.read())
-  disk_file.close()
-  database = geoip2.database.Reader(disk_filename, mode=MODE_MEMORY)
-  os.remove(disk_filename)
-  return database
+  with open(disk_filename, 'wb') as disk_file:
+    disk_file.write(f.read())
+    disk_file.close()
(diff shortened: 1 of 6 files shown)


pipeline/metadata/maxmind.py
```diff
@@ -28,12 +28,12 @@ def _maxmind_reader(filepath: str) -> geoip2.database.Reader:
...
   disk_filename = os.path.join('/tmp', os.path.basename(filepath))
-  disk_file = open(disk_filename, 'wb')
-  disk_file.write(f.read())
-  disk_file.close()
-  database = geoip2.database.Reader(disk_filename, mode=MODE_MEMORY)
-  os.remove(disk_filename)
-  return database
+  with open(disk_filename, 'wb') as disk_file:
+    disk_file.write(f.read())
+    disk_file.close()
```
avirkud, line 33: The close can be removed and the rest can be moved out of the `with` block.

avirkud (approved):

pipeline/metadata/maxmind.py
ohnorobo: done

PR #18, opened 2021-04-30
update blockpage matcher interface to include signatures

ohnorobo:
Let me know what you think of this interface/final columns.

### DIFF
--- a/pipeline/beam_tables.py
+++ b/pipeline/beam_tables.py
@@ -69,7 +69,7 @@ SCAN_BIGQUERY_SCHEMA = {
     'measurement_id': ('string', 'nullable'),
     'source': ('string', 'nullable'),
     'blockpage': ('boolean', 'nullable'),
-    'blockpage_signature': ('string', 'nullable'),
+    'page_signature': ('string', 'nullable'),
 
     # Column filled in all tables
     'received_status': ('string', 'nullable'),
@@ -549,7 +549,7 @@ class FlattenMeasurement(beam.DoFn):
     if anomaly:  # check response for blockpage
       blockpage, signature = self.blockpage_matcher.match_page(received['body'])
       row['blockpage'] = blockpage
-      row['blockpage_signature'] = signature
+      row['page_signature'] = signature
 
     tls = received.get('tls', None)
     if tls:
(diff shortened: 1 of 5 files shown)


ramakrishnansr (approved):
Just thinking about this now, but I'm wondering whether `re.compile` handles Unicode characters well? Have seen that Python's default regex library is not very goot at matching unicode (https://docs.python.org/3/library/re.html). Is there a need to add `re.UNICODE` to the compile? The docs suggest using this third party regex library for a more thorough regex matching of Unicode characters (https://pypi.org/project/regex/).

pipeline/beam_tables.py
```diff
@@ -544,12 +544,12 @@ def _parse_received_data(self, received: Union[str, Dict[str, Any]],
...
             _parse_received_headers(received.get('headers', {})),
-        'blockpage':
-            None,
     }
 
     if anomaly:  # check response for blockpage
-      row['blockpage'] = self.blockpage_matcher.match_page(received['body'])
+      blockpage, signature = self.blockpage_matcher.match_page(received['body'])
+      row['blockpage'] = blockpage
+      row['blockpage_signature'] = signature
```
fortuna, line 552: Let's call this page_signature. We don't need to tie it to "blockpage". It could be any well known page (like the Cloudflare ddos protection page). In fact, in some cases the signature is already not for a block page.

fortuna (approved):
Please rename the field

ohnorobo:
> Just thinking about this now, but I'm wondering whether `re.compile` handles Unicode characters well? Have seen that Python's default regex library is not very goot at matching unicode (https://docs.python.org/3/library/re.html). Is there a need to add `re.UNICODE` to the compile? The docs suggest using this third party regex library for a more thorough regex matching of Unicode characters (https://pypi.org/project/regex/).

@avirkud did add a unicode test case [here](https://github.com/censoredplanet/censoredplanet-analysis/blob/1536706418fa1830b5186e00e84a0d83de37ce99/pipeline/test_beam_tables.py#L771) with a bunch of special characters. But it would be good to add another testing blockpages in persian or chinese. I'll look into that separately.

pipeline/beam_tables.py
ohnorobo: done

PR #22, opened 2021-06-03
beam jobs fail when they create zero rows

ohnorobo:
Currently we're mostly bypassing individual bad/unparsable rows in json files. This is generally fine, but if the whole file is bad (or, in our current situation, the whole file is in a new format) then the beam job can succeed even though we've failed entirely.

I'm testing this in the e2e test because testing things like raised errors using beam's unit testing framework doesn't seem to work.

Ran a test job [here](https://console.cloud.google.com/dataflow/jobs/us-west1/2021-06-03_09_47_43-17483800278137902953;step=read%20file%20gs:?project=firehook-censoredplanet) and saw that it failed.

### DIFF
--- a/pipeline/beam_tables.py
+++ b/pipeline/beam_tables.py
@@ -700,6 +700,19 @@ def get_table_name(dataset_name: str, scan_type: str,
   return f'{dataset_name}.{scan_type}_{base_table_name}'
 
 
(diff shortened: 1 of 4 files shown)


avirkud (approved):
Looks good!

PR #23, opened 2021-06-04
add parsing for hyperquack v2 format

ohnorobo:
Overall output format changes so that v1 and v2 can use the same output format:
- `blocked` field renamed to `anomaly` to match v2 names
- `is_control` field added
- `controls_failed` field added (and calculated for v1)
- `retries` field removed on the grounds that it's just not that important
- `domain` field changed in v1 to populate from the `Sent` field whenever possible (meaning we get the control domain instead of the overall testing domain for control probes.)
- `is_control` field in v1 populated from the domain in the sent field, or else from the retry index when the Sent field is missing.
- deleted the output `sent` field
- deleted `fail_sanity` field. This is superseded by `controls_failed`

Updated base tables [doc](https://github.com/censoredplanet/censoredplanet-analysis/blob/v2/docs/base_tables.md).

Possible things to change:
- For the incorrectly formatted `sent` in some discard tests I'm currently getting the correct domain. Should I get no domain instead?
- This is also an easy time to change any other column names you'd like, since we have to do a full backfill anyway.

Things I'm not doing in this PR:
- Hyperquack v2 doesn't populate the error field with the same `Incorrect web response: etc` content that v1 does, so currently those entries have no errors populated. I'll work on making those match in a future change.
- Dealing with the new location data from v2.

Testing:
- ran some beam jobs test parsing a week of v1 data and week of v2 data into side tables.
- ran `python3 -m pipeline.run_beam_tables --env=user --scan_type=all --full --start_date=2021-03-01 --end_date=2021-07-01` to test parsing both types of data at once


docs/base_tables.md
```diff
@@ -92,12 +66,72 @@ The json data is processed into a flat table format which looks like this.
 | **Analysis**              |
 |                           |
 | success                   | BOOLEAN      | Did the individual roundtrip measurement succeed? |
-| blocked                   | BOOLEAN      | Was interference detected in the overall measurement? |
-| fail_sanity               | BOOLEAN      | Was the ip being tested malfunctioning/down? |
+| anomaly                   | BOOLEAN      | Was interference detected in the overall measurement? |
+| fail_sanity               | BOOLEAN      | Was the ip being tested malfunctioning/down? </br> :warning: only present in v1 data |
```
ramakrishnansr, line 70: This and control_failed can be merged.

pipeline/beam_tables.py
```diff
@@ -49,13 +49,14 @@
...
-    'sent': ('string', 'nullable'),
+    'sent': ('string', 'nullable'),  # v1 only
     'error': ('string', 'nullable'),
-    'blocked': ('boolean', 'nullable'),
+    'anomaly': ('boolean', 'nullable'),
     'success': ('boolean', 'nullable'),
-    'fail_sanity': ('boolean', 'nullable'),
+    'fail_sanity': ('boolean', 'nullable'),  # v1 only
     'stateful_block': ('boolean', 'nullable'),
+    'is_control': ('boolean', 'nullable'),
```
ramakrishnansr, line 58: Is this calculated for v1 also?

ramakrishnansr (commented):
One thing I would like to point out is that some changes to v2 were made gradually, so for your tests, you may want to take one file from the beginning of March and one file from Mid-June to ensure that both formats work well.

docs/base_tables.md
ohnorobo: done

pipeline/beam_tables.py
ohnorobo: yes

pipeline/metadata/flatten.py
```diff
@@ -68,6 +82,55 @@ def source_from_filename(filepath: str) -> str:
...
+    scan: a loaded json object containing the parsed content of the line
+
+  Returns: bool
+  """
+  for result in scan.get('Results', []):
+    actual_domain = _extract_domain_from_sent_field(result['Sent'])
+    is_control = _is_control_url(actual_domain)
+    if is_control and not result['Success']:
+      return True
+  return False
```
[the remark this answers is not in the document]
ohnorobo: deleted

ohnorobo:
> * `controls_failed` and `failed_sanity` are interchangeable, I would suggest keeping one and removing the other.

done

> * I think the `sent` field can be removed now.

done

> * Yes, we do need to calculate what's the sent field using the number of retries. In some cases where the control measurement also fails due to a network timeout, it was not populated in the `sent` field in v1.

done, although I'm not adding back in the control domain for missing sent fields, only the regular test domains.

That means a non-control probe with `sent` missing will look like
"domain": "test_url.com" <- extracted from the overall measurement domain
"is_control": False

and for a control probe
"domain": "" <- missing, but we don't care much about the actual url as long as we know it's a control
"is_control": True

> * I'm not sure about the problem you're mentioning about the discard tests, can you give me an example?

Some of the discard tests from 2018/2019 use the second pattern listed [here](https://github.com/censoredplanet/censoredplanet-analysis/blob/0c1592ace0dcd1dc4415deb308c585433f4319e6/pipeline/metadata/flatten.py#L32) instead of the first one. Currently I'm extracting the intended url, but it might be preferable to do something else like drop the probes as invalid.

> * Yes, it should only be true when all control failed. A control failing immediately after the test could be due to stateful blocking.

done

ohnorobo:
> One thing I would like to point out is that some changes to v2 were made gradually, so for your tests, you may want to take one file from the beginning of March and one file from Mid-June to ensure that both formats work well.

I was worried this might mean there was some mixed format data, so I ran a test parse into a side table over these dates

`python3 -m pipeline.run_beam_tables --env=user --scan_type=all --full --start_date=2021-03-01 --end_date=2021-07-01`

All four of them succeeded, (any uncaught exceptions when parsing a single line cause the whole pipeline to fail.) so I think we're good. I had already added both v1 and v2 data example files in the e2e test.

PR #28, opened 2021-06-29
classify content mismatch for hyperquack v2 and match some new errors

ohnorobo:
This is the simple/hacky way to fix the different content in the error field for hyperquack v2. The correct way will be to move this logic into the beam pipeline itself.

Two major changes:
- Hyperquack v2 doesn't include the "Incorrect web response *" errors that v1 does. But it does include a `template_mismatch` field. This uses that field to separate out content problems.
- There are a few new errors that shows up in v2 vs v1. This includes those in the enum.

One minor change:
- Suppressing a new warning that appeared in the latest version of pylint.

New errors are
- Get "https://[IP]:[PORT]/": http: server closed idle connection (Client.Timeout exceeded while awaiting headers)
- Get "https://[IP]:[PORT]/": net/http: TLS handshake timeout
- Get "https://[IP]:[PORT]/": net/http: TLS handshake timeout (Client.Timeout exceeded while awaiting headers)
- Get "https://[IP]:[PORT]/": read tcp [IP]:[PORT]->[IP]:[PORT]: read: connection reset by peer (Client.Timeout exceeded while awaiting headers)

which map to
- `read/http.truncated_response`
- `tls/timeout`
- `tls/timeout`
- `read/tcp.reset`

respectively

Tested: ran queries over the prod data
- saw data transition form a bunch of different `content/*` outcomes to just `content/template_mismatch`
- there are no `unknown` errors in the final dataset

4GHrtr7tDXAYNq5

Unexpectedly there were some `content/template_mismatch` errors surfaced in the v1 data as well. This can only happen when `Success` (the v1 (and boolean opposite) field name for `matches_template` in v2 is false but `error` is null/empty). Looking at the data I think there is probably a v1 error where sometimes the "Incorrect web response *" data in `error` wasn't being populated, so this is actually a reasonable thing to do. It's also only a small percentage of the data.

### DIFF
--- a/docs/merged_reduced_scans_table.md
+++ b/docs/merged_reduced_scans_table.md
@@ -117,7 +117,10 @@ Mismatch Errors are used when the connection is successful, but the content rece
 |                         |
 | **Mismatch Errors**     | The connection completed successfully, but the content returned didn't match the content expected for the domain. |
 |                         |
+| HyperQuack v1 only      |
 | response_mismatch       | Received a different response from the one expected. </br> For Discard no response is expected and any response is a mismatch, </br> for Echo a mirrored response is expected and anything else is a mismatch. |
 | status_mismatch         | The HTTP status code didn't match, eg. `403` instead of `200` |
 | body_mismatch           | The HTTP body didn't match, potentially a blockpage |
 | tls_mismatch            | An element of the TLS connection (certificate, cipher suite, or TLS version) didn't match |
+| HyperQuack v1 and v2    |
+| template_mismatch       | Some element of the response did not match the expected template |
--- a/table/queries/merged_reduced_scans.sql
+++ b/table/queries/merged_reduced_scans.sql
@@ -24,13 +24,17 @@ CREATE TEMP FUNCTION CleanError(error STRING) AS (
 # Classify all errors into a small set of enums
 #
 # Input is a nullable error string from the raw data
-# Source is one of "ECHO", "Discard, "HTTP", "HTTPS"
+# Source is one of "ECHO", "DISCARD, "HTTP", "HTTPS"
 #
 # Output is a string of the format "stage/outcome"
 # Documentation of this enum is at
 # https://github.com/censoredplanet/censoredplanet-analysis/blob/master/docs/tables.md#outcome-classification
-CREATE TEMP FUNCTION ClassifyError(error STRING, source STRING) AS (
+CREATE TEMP FUNCTION ClassifyError(error STRING, source STRING, template_match BOOL) AS (
   CASE
+    # Content mismatch for hyperquack v2 which doesn't write
+    # content verification failures in the error field.
+    WHEN (NOT template_match AND (error is NULL OR error = "")) then "content/template_mismatch"
+
     # Success
     WHEN (error is NULL OR error = "") then "complete/success"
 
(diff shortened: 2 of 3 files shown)


docs/merged_reduced_scans_table.md
```diff
@@ -117,7 +117,10 @@ Mismatch Errors are used when the connection is successful, but the content rece
 |                         |
 | **Mismatch Errors**     | The connection completed successfully, but the content returned didn't match the content expected for the domain. |
 |                         |
+| HyperQuack v1 only      |
 | response_mismatch       | Received a different response from the one expected. </br> For Discard no response is expected and any response is a mismatch, </br> for Echo a mirrored response is expected and anything else is a mismatch. |
 | status_mismatch         | The HTTP status code didn't match, eg. `403` instead of `200` |
 | body_mismatch           | The HTTP body didn't match, potentially a blockpage |
 | tls_mismatch            | An element of the TLS connection (certificate, cipher suite, or TLS version) didn't match |
+| HyperQuack v1 and v2    |
+| template_mismatch       | Some element of the response did not match the expected template |
```
fortuna, line 126: You might even mention the field you use to calculate this.

table/queries/merged_reduced_scans.sql
```diff
@@ -57,6 +61,7 @@ CREATE TEMP FUNCTION ClassifyError(error STRING, source STRING) AS (
     WHEN ENDS_WITH(error, "readLoopPeekFailLocked: <nil>") THEN "tls/tls.failed"
     WHEN ENDS_WITH(error, "missing ServerKeyExchange message") THEN "tls/tls.failed"
     WHEN ENDS_WITH(error, "no mutual cipher suite") THEN "tls/tls.failed"
+    WHEN REGEXP_CONTAINS(error, "TLS handshake timeout") THEN "tls/timeout"
```
fortuna, line 64: Why is this s regex now?
Isn't there a simple string contains?
Same for the other ones

ohnorobo: Bigquery only has REGEXP_CONTAINS and REGEXP_MATCH.

docs/merged_reduced_scans_table.md
ohnorobo: Mentioning how the json->bigquery field mapping is done for all these fields would be too long and distracting. I'm thinking of this documentation as intended for someone who is only using the bq tables and doesn't need to know much about the json.

table/queries/merged_reduced_scans.sql
ohnorobo: ENDS_WITH/STARTS_WITH are used when the string only appears at the end/beginning of an error. If it appears in the middle we have to use REGEXP_CONTAINS.

fortuna: There's also STRPOS. It may be worth comparing the performance, since BigQuery is a huge cost, but I'm fine either way.

fortuna (approved):

table/queries/merged_reduced_scans.sql
ohnorobo: I tried timing it and the strpos version took 3% longer. So probably the query planner is doing something smart and they're basically equivalent.

PR #30, opened 2021-07-07
normalize all control domains

ohnorobo:
[hyperquack v2 measurements](https://github.com/censoredplanet/censoredplanet-analysis/pull/23) generate a bunch of control domains like control-123456789.com which don't add unique info and clutter up the dashboard. This normalizes all control domains to the same name.

tested: ran on prod tables

### DIFF
--- a/table/queries/merged_reduced_scans.sql
+++ b/table/queries/merged_reduced_scans.sql
@@ -127,7 +127,7 @@ WITH
   AllScans AS (
   SELECT
     date,
-    domain,
+    IF(is_control, "CONTROL", domain) as domain,
     "DISCARD" AS source,
     country,
     netblock,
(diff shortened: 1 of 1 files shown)


fortuna (approved):
This will normalize v2 measurements only. Could you add normalization for v1 too? It can be on another PR.

ohnorobo:
> This will normalize v2 measurements only. Could you add normalization for v1 too? It can be on another PR.

The `is_control` field is calculated for v1 data [here](https://github.com/censoredplanet/censoredplanet-analysis/blob/aaa81100d5e5a883b30fbc379863fb046dbba945/pipeline/metadata/flatten.py#L225) and has been backfilled. So this works for both v1 and v2.

PR #32, opened 2021-07-20
add failed controls info to table

ohnorobo:
Would it be clearer for dashboard purposes to reverse this field and instead have a `controls_succeeded` field?


ramakrishnansr (approved):

ramakrishnansr:
> Would it be clearer for dashboard purposes to reverse this field and instead have a `controls_succeeded` field?

I am not sure, I think the primary use would be to filter out by `controls_failed`, so it might be best to keep it as is.

PR #33, opened 2021-07-23
add a specific outcome for blockpages

ohnorobo:
This adds blockpage matches as a possible outcome. Blockpages are a subset of content mismatches but provide much clearer evidence of blocking, so they're worth calling out individually.

This doesn't do anything to show which specific blockpage matched since that would require displaying a new field in the dashboard, whereas this change just shows up automatically. We can think more about if/how to do that in the future.

Here's an [dashboard example](https://datastudio.google.com/s/saNPMZcHVco) of how this looks investigating blogger blocking in Iran on HTTP.
568GVX7jE9M3KMP

### DIFF
--- a/docs/merged_reduced_scans_table.md
+++ b/docs/merged_reduced_scans_table.md
@@ -124,3 +124,4 @@ Mismatch Errors are used when the connection is successful, but the content rece
 | tls_mismatch            | An element of the TLS connection (certificate, cipher suite, or TLS version) didn't match |
 | HyperQuack v1 and v2    |
 | template_mismatch       | Some element of the response did not match the expected template |
+| blockpage               | The response was unexpected and matched a [known blockpage]((https://github.com/censoredplanet/censoredplanet-analysis/blob/master/pipeline/metadata/data/blockpage_signatures.json)) |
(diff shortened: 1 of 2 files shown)


fortuna (approved):
Very useful. I wonder how it would look like if we showed the fingerprint id in the outcome, like `content/blockpage(id)`

ohnorobo:
I looked into making a [side dashboard](https://datastudio.google.com/c/u/0/reporting/aa1bcee7-6c87-481c-ba60-60920414ae22) which shows blockpage signatures.

I'm not a huge fan of it because even a few blockpages overflow the series label display for the timecline chart and make it hard to work with
C6rk256o8vupNFa

But feel free to play around with it.

This also reminds me of my other objection to directly showing blockpage signatures, which is that end users don't really care about the difference between `b_nat_ir_national_1`, `b_nat_ir_national_2`, `b_nat_ir_national_3`. We need to think about a better user-friendly way to unify and display that info before putting it in the dashboard.

fortuna:
Talked offline. I wouldn't worry about the label overflow: once you select an AS, you only get one blockpage.

I'd like to keep the ids, perhaps reformatted as something like `content/blockpage/b_nat_ir_national_3`, `content/blockpage[b_nat_ir_national_3]` or `content/blockpage(b_nat_ir_national_3)`.

We can get interesting insights. For example, a private company network in Iran returns a Cisco block page, indicating it's local censorship, not national. If we find it overwhelming later, we can remove it.

PR #35, opened 2021-07-26
classify v2 outcomes

ohnorobo:
Classify the new outcome strings from hyperquack v2 in the same way we classified v1 strings.

These started appearing after the fix in https://github.com/censoredplanet/hyperquackv2/pull/52 was released.

unknown errors seen in the diagnostic dashboard:
8UtTgCtiYjwSvgz

### DIFF
--- a/table/queries/merged_reduced_scans.sql
+++ b/table/queries/merged_reduced_scans.sql
@@ -90,6 +90,7 @@ CREATE TEMP FUNCTION ClassifyError(error STRING, source STRING, template_match B
 
     # Content verification failures
     WHEN (source = "ECHO" AND ENDS_WITH(error, "EOF")) THEN "content/response_mismatch" # Echo
+    # Hyperquack v1 errors
     WHEN (error = "Incorrect echo response") THEN "content/response_mismatch" # Echo
     WHEN (error = "Received response") THEN "content/response_mismatch" # Discard
     WHEN (error = "Incorrect web response: status lines don't match") THEN "content/status_mismatch" # HTTP/S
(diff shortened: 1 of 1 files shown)


table/queries/merged_reduced_scans.sql
```diff
@@ -90,13 +90,22 @@ CREATE TEMP FUNCTION ClassifyError(error STRING, source STRING, template_match B
...
     WHEN (error = "Received response") THEN "content/response_mismatch" # Discard
     WHEN (error = "Incorrect web response: status lines don't match") THEN "content/status_mismatch" # HTTP/S
     WHEN (error = "Incorrect web response: bodies don't match") THEN "content/body_mismatch" # HTTP/S
     WHEN (error = "Incorrect web response: certificates don't match") THEN "content/tls_mismatch" # HTTPS
     WHEN (error = "Incorrect web response: cipher suites don't match") THEN "content/tls_mismatch" # HTTPS
     WHEN (error = "Incorrect web response: TLS versions don't match") THEN "content/tls_mismatch" # HTTPS
+    # Hyperquack v2 errors
+    WHEN (error = "echo response does not match echo request") THEN "content/response_mismatch" # Echo
+    WHEN (error = "discard response is not empty") THEN "content/response_mismatch" # Discard
+    WHEN (error = "Status lines does not match") THEN "content/status_mismatch" # HTTP/S
```
ramakrishnansr, line 104: One thing we may want to make clear either in the documentation or somewhere in the pipeline is that Hyperquackv2 (and v1) only outputs the first mismatch it finds. Cases of "content/status_mismatch" frequently also have a body mismatch. This also means that the TLS mismatch cases are especially interesting as they possibly show cases of transparent TLS interception for certain domains.

ramakrishnansr (approved):

table/queries/merged_reduced_scans.sql
ohnorobo: Good point, probably the best place to put that would be in the docs. I'll have to think about how best to explain.

table/queries/merged_reduced_scans.sql
```diff
@@ -90,13 +90,22 @@ CREATE TEMP FUNCTION ClassifyError(error STRING, source STRING, template_match B
...
+    WHEN (error = "Bodies do not match") THEN "content/body_mismatch" # HTTP/S
+    WHEN (error = "Certificates do not match") THEN "content/tls_mismatch" # HTTPS
+    WHEN (error = "Cipher suites do not match") THEN "content/tls_mismatch" # HTTPS
+    WHEN (error = "TLS versions do not match") THEN "content/tls_mismatch" # HTTPS
 
     # Unknown errors
     ELSE "unknown/unknown"
```
fortuna, line 111: Perhaps we should differentiate missing error from uncategorized error.
When I saw this in the dashboard I thought we didn't know what the error was.

Also, outputting the error when uncategorized would help the dashboard degrade gracefully, enabling the user to analyze outcomes that we haven't seen before.

PR #38, opened 2021-07-27
output blockpage signatures

ohnorobo:
This will output outcomes like `content/blockpage:b_nat_ir_national_3`

### DIFF
--- a/table/queries/merged_reduced_scans.sql
+++ b/table/queries/merged_reduced_scans.sql
@@ -29,9 +29,13 @@ CREATE TEMP FUNCTION CleanError(error STRING) AS (
 # Output is a string of the format "stage/outcome"
 # Documentation of this enum is at
 # https://github.com/censoredplanet/censoredplanet-analysis/blob/master/docs/tables.md#outcome-classification
-CREATE TEMP FUNCTION ClassifyError(error STRING, source STRING, template_match BOOL, blockpage_match BOOL) AS (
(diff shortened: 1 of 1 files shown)


table/queries/merged_reduced_scans.sql
```diff
@@ -29,9 +29,13 @@ CREATE TEMP FUNCTION CleanError(error STRING) AS (
...
 # https://github.com/censoredplanet/censoredplanet-analysis/blob/master/docs/tables.md#outcome-classification
-CREATE TEMP FUNCTION ClassifyError(error STRING, source STRING, template_match BOOL, blockpage_match BOOL) AS (
+CREATE TEMP FUNCTION ClassifyError(error STRING,
+                                   source STRING,
+                                   template_match BOOL,
+                                   blockpage_match BOOL,
+                                   blockpage_signature STRING) AS (
   CASE
-    WHEN blockpage_match then "content/blockpage"
+    WHEN blockpage_match then CONCAT("content/blockpage:", blockpage_signature)
```
fortuna, line 38: You need to SQL-escape the signature.

Also, `blockpage_id` would probably be more accurate, since it's not the signature itself.

ohnorobo: Changed to `blockpage_id`.

Sql escaping discussed offline, not relevant here.

fortuna (approved):

PR #45, opened 2021-08-11
turn off maxmind data additions

ohnorobo:
Turn off maxmind data use and split out + disable the maxmind data test.

### DIFF
--- a/pipeline/beam_tables.py
+++ b/pipeline/beam_tables.py
@@ -988,7 +988,8 @@ class ScanDataBeamPipelineRunner():
     caida_ip_metadata_db = self.caida_ip_metadata_class(
         datetime.date.fromisoformat(date), self.caida_ip_metadata_bucket_folder,
         True)
-    maxmind_db = self.maxmind_class(self.maxmind_bucket_folder)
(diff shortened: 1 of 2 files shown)


pipeline/beam_tables.py
```diff
@@ -1004,10 +1004,11 @@ def _add_ip_metadata(self, date: str,
             'as_class': as_type,
             'country': country,
         }
-        if not metadata_values['country']:  # try Maxmind
-          (netblock, asn, as_name, as_full_name, as_type,
-           country) = maxmind_db.lookup(ip)
-          metadata_values['country'] = country
+        # Turning off maxmind data for now.
```
fortuna, line 1007: I'd disable the maxmind_db construction above too to save resources and time.

fortuna (approved):

pipeline/beam_tables.py
fortuna: Ideally we should just inject a IP->country mapping object, rather than having to know about the implementations and fallback. No need to make that change though in this PR though.

ohnorobo: disabled

PR #46, opened 2021-08-12
Add Satellite measurements from control resolvers

avirkud:
Load Satellite measurements from control resolvers into the tables.
- `responses_control.json` added to input files
- Function to process response file format

Changes to BQ tables:
- `has_type_a`: new field (Satellite only). v2 data indicates if we receive a type A RR
- `controls_failed`: existing field from Hyperquack tables - this is filled based on the `passed_control` field in the raw Satellite data (refers to measurements to the control domain). 
- This resolves the issue where `anomaly` was always set to False in the tables


pipeline/metadata/flatten.py
```diff
@@ -366,13 +370,52 @@ def _process_satellite_v2(self, scan: Any,
...
+        'domain': scan['test_url'],
+        'category': self.category_matcher.match_url(scan['test_url']),
+        'ip': scan['vp'],
+        'anomaly': None,
+        'success': not scan['connect_error'],
+        'controls_failed': not scan['passed_control'],
+        'measurement_id': random_measurement_id
+    }
+    responses = scan.get('response', [])
+    if responses:
```
ohnorobo, line 402: change `resp` -> `response` in this whole section, just for easier reading.

pipeline/metadata/flatten.py
```diff
@@ -366,13 +370,52 @@ def _process_satellite_v2(self, scan: Any,
...
+      row['rcode'] = [str(resp['rcode']) for resp in responses]
+      row['error'] = ' | '.join(
+          [resp['error'] for resp in responses if resp['error']])
+      for resp in responses:
+        if resp['url'] == row['domain']:
+          # Check response for test domain
+          if resp['rcode'] == 0 and resp['has_type_a']:
+            # Valid answers
+            row['has_type_a'] = True
+            for ip in resp['response']:
```
ohnorobo, line 415: If there are multiple ips in `resp['response']` won't this just add the first and yield immediately instead of iterating over all of them?

pipeline/metadata/flatten.py
```diff
@@ -366,13 +370,52 @@ def _process_satellite_v2(self, scan: Any,
...
         'success': not scan['connect_error'],
+        'controls_failed': not scan['passed_control'],
         'received': None,
         'rcode': [],
         'measurement_id': random_measurement_id
     }
     received_ips = scan.get('response')
     yield from self._process_received_ips(row, received_ips)
 
+  def _process_satellite_v2_responses(
```
ohnorobo, line 381: Add a test for this function and also add an example responses_control.json file to the e2e test.

ohnorobo (changes requested):

pipeline/metadata/flatten.py
avirkud: Done

pipeline/metadata/flatten.py
avirkud: The flattened rows for Satellite each have one answer IP for tagging purposes (this is similar to `_process_received_ips`).

pipeline/metadata/flatten.py
avirkud: Updated e2e test files and unit tests (`test_flattenmeasurement_dns`, `test_process_satellite_v2`)

pipeline/metadata/flatten.py
ohnorobo: In that case I would just write it as

```
ip = response['response'][0]
```

so the intent is clearer

avirkud: Just to clarify, `response['response']` can have multiple ips, but this function is flattening the measurement so that each yielded row has one answer IP. 

`test_flatten.py` has an example of this in `test_flattenmeasurement_dns`. The `response['response']` field has three answer IPs, resulting in three output rows.
```
{
          "vp": "8.8.4.4",
          "test_url": "login.live.com",
          "response": [
            ...
            {
              "url": "login.live.com",
              "has_type_a": true,
              "response": [
                "40.126.31.135",
                "40.126.31.8",
                "40.126.31.6"
              ],
              ...
            },
            ...
          ],
          ...
}
```

pipeline/metadata/test_flatten.py
```diff
@@ -1109,9 +1156,58 @@ def test_flattenmeasurement_dns(self) -> None:
...
+        'start_time': '2021-03-15T20:02:06.996014164-04:00',
+        'end_time': '2021-03-15T20:02:07.015386713-04:00',
+        'error': None,
+        'anomaly': None,
+        'success': True,
+        'controls_failed': False,
+        'received': {'ip': '40.126.31.6'},
+        'rcode': ['0', '0', '0'],
+        'measurement_id': '',
+        'has_type_a': True
```
ohnorobo, line 1210: I can't comment on the exact line, but at line 1121 can you change the code

```
for row in rows:
  row['measurement_id'] = ''
  result.append(row.copy())
```

To something like

```
results.append(list(rows))
for result in results:
  result['measurement_id'] = ''
```

I'm worried the .copy() in that code may be hiding a bug where we're yielding the same row object multiple times while overwriting its ip field.

pipeline/metadata/flatten.py
```diff
@@ -366,13 +370,56 @@ def _process_satellite_v2(self, scan: Any,
...
+      row['error'] = ' | '.join(errors) if errors else None
+      for response in responses:
+        if response['url'] == row['domain']:
+          # Check response for test domain
+          if response['rcode'] == 0 and response['has_type_a']:
+            # Valid answers
+            row['has_type_a'] = True
+            for ip in response['response']:
+              row['received'] = {'ip': ip}
+              yield row
```
ohnorobo, line 421: I'm a bit worried that this yield is emitting the same row object multiple times while just overwriting its ip field. You may need to make a copy of the row before emitting here, see the comment below on the test.

pipeline/metadata/test_flatten.py
avirkud: Done

pipeline/metadata/flatten.py
avirkud: This was the issue! `_process_received_ips` also had the same problem. I've updated these to yield a copy of the row.

ohnorobo (approved):

PR #51, opened 2021-08-19
refactor ip metadata classes

ohnorobo:
Refactor the addition of ip metadata out into its own class `IpMetadataChooser` which contains the class/folder objects it needs and can be lazy-initialized in the workers when needed.

I've also removed `IpMetadataInterface` which was being used for both caida/maxmind and instead given each source of metadata (currently caida/maxmind/dbip) its own personal interface which is then used to define the fake. It never made much sense to expect the different sources to have the same interface anyway, since they return different kinds of values.

This is a pure refactor and should not change any behavior/data output.


pipeline/beam_tables.py
```diff
@@ -849,40 +851,22 @@ class ScanDataBeamPipelineRunner():
...
       bucket: gcs bucket name
       staging_location: gcs bucket name, used for staging beam data
       temp_location: gcs bucket name, used for temp beam data
-      caida_ip_metadata_class: an IpMetadataInterface subclass (class, not instance)
-      caida_ip_metadata_bucket_folder: gcs folder with CAIDA ip metadata files
-      maxmind_class: an IpMetadataInterface subclass (class, not instance)
-      maxmind_bucket_folder: gcs folder with maxmind files
-      dbip_class: a DbipMetadata class (class, not instance)
-      dbip_bucket_folder: gcs folder with dbip files
+      metadata_prod_values: values used to initialize an IpMetadataChooser
```
fortuna, line 863: There's no need for a `IpMetadataValues` and a `IpMetadataChooser`.
Just pass a `IpMetadataChooser`.
I imagine that the problem is that it has to get initialized in each worker. You can do lazy initialization on the first lookup.
Alternatively, pass a factory function that will return an initialized `IpMetadataChooser` given a date. The `IpMetadataValues` is an implementation detail that this code shouldn't need to know.

pipeline/ip_metadata_values.py
```diff
@@ -0,0 +1,51 @@
+"""A module to hold a bunch of values needed to init an IPMetadataChooser."""
```
fortuna, line 1: This file can be deleted. Pass a factory function to the pipeline or make the chooser know how to initialize itself.

pipeline/metadata/caida_ip_metadata.py
```diff
@@ -37,6 +35,24 @@
...
+      self,
+      date: datetime.date,
+      cloud_data_location: str,
+      allow_previous_day: bool,
+  ) -> None:
+    pass
+
+  def lookup(
+      self, ip: str
+  ) -> Tuple[Optional[str], int, Optional[str], Optional[str], Optional[str],
```
fortuna, line 51: These tuples are hard to follow. Perhaps define a named tuple?
https://docs.python.org/3/library/collections.html#collections.namedtuple

pipeline/metadata/ip_metadata_chooser.py
```diff
@@ -0,0 +1,51 @@
...
+
+from pipeline.metadata.flatten import Row
+from pipeline.ip_metadata_values import IpMetadataValuesInterface
+
+
+class IpMetadataChooser():
+  """Business logic for selecting IP metadata values from a number of sources"""
+
+  def __init__(self, date: datetime.date,
+               prod_values: IpMetadataValuesInterface) -> None:
```
fortuna, line 13: Pass the dbip, maxmind and caida objects here instead. No need to pass the folders.

pipeline/metadata/dbip.py
```diff
@@ -11,7 +11,17 @@
...
 
 
-class DbipMetadata():
+class DbipIpMetadataInterface:
+  """Interface for an CAIDA IP Metadata lookup database."""
+
+  def __init__(self, dbip_folder: str) -> None:
+    pass
+
+  def lookup(self, ip: str) -> Tuple[Optional[str], Optional[int]]:
```
fortuna, line 20: Same: define a named tuple

fortuna (changes requested):

pipeline/beam_tables.py
```diff
@@ -31,6 +31,8 @@
 from pipeline.lookup_country_code import country_name_to_code
 from pipeline.metadata.flatten import Row
 from pipeline.metadata import flatten
+from pipeline.ip_metadata_values import IpMetadataValuesInterface
```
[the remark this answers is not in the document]
ohnorobo: I'm using -`Interface` because the non-interface name is already taken for the actual class implementation

ex for caida:
`CaidaIpMetadataInterface` / `CaidaIpMetadata` (actual implementation) / `FakeCaidaIpMetadata` (fake for testing)

What pattern should I change these three to?

pipeline/beam_tables.py
ohnorobo: made IpMetadataChooser contain the values and have a lazy init option.

pipeline/ip_metadata_values.py
ohnorobo: done

pipeline/metadata/caida_ip_metadata.py
ohnorobo: added named tuples are return types

pipeline/metadata/ip_metadata_chooser.py
ohnorobo: I can't pass the objects because they can't be initialized yet at this point.

I've switched to passing in the classes/folders and having a lazy init that takes the date.

pipeline/metadata/dbip.py
ohnorobo: done

pipeline/beam_tables.py
```diff
@@ -1016,44 +998,12 @@ def _add_ip_metadata(self, date: str,
       Tuples (DateIpKey, metadata_dict)
       where metadata_dict is a row Dict[column_name, values]
     """
-    caida_ip_metadata_db = self.caida_ip_metadata_class(
-        datetime.date.fromisoformat(date), self.caida_ip_metadata_bucket_folder,
-        True)
-    # TODO turn back on when using maxmind again.
-    # maxmind_db = self.maxmind_class(self.maxmind_bucket_folder)
-    dbip_db = self.dbip_class(self.dbip_bucket_folder)
+    self.ip_metadata_chooser.fully_init(datetime.date.fromisoformat(date))
```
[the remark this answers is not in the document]
fortuna: Something like:
```py
ip_metadata_chooser = self.create_metadata_chooser(date)
```

pipeline/metadata/ip_metadata_chooser.py
```diff
@@ -0,0 +1,110 @@
...
+from pipeline.metadata import dbip
+from pipeline.metadata import maxmind
+from pipeline.metadata.flatten import Row
+
+
+# pylint: disable=too-many-instance-attributes
+class IpMetadataChooser():
+  """Business logic for selecting IP metadata values from a number of sources"""
+
+  def __init__(self, caida_ip_metadata_class: Type[
```
fortuna, line 16: Instead of this class, provide a function that returns a factory function to avoid the partial initialization.

Create a new `IpMetadataChooser` that takes in the constructor:
- the mm_reader for dbip (and maxmind).
- the asn_db, as_to_type_map and as_to_org_map from Caida

This way you don't need all the different classes and interfaces. You are not gaining much from them. Just pass fake mm_readers and dicts in the tests.

pipeline/metadata/caida_ip_metadata.py
```diff
@@ -36,6 +34,30 @@
...
+# ex: ("1.0.0.1/24", 13335, "CLOUDFLARENET", "Cloudflare Inc.", "Content", "US")
+CaidaReturnValues = NamedTuple('CaidaReturnValues',
+                               [('netblock', Optional[str]), ('asn', int),
+                                ('as_name', Optional[str]),
+                                ('as_full_name', Optional[str]),
+                                ('as_type', Optional[str]),
+                                ('country', Optional[str])])
+
+
+class CaidaIpMetadataInterface:
```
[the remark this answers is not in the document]
fortuna: But seem my other comment below. I believe the interface and class can be deleted.

fortuna (changes requested):
The current architecture can be simplified.
- Get rid of all those classes and interfaces.
- Have one metadata chooser that takes as input the DBIP mm_reader and Caida dictionaries. Those can be faked for testing. No need to know about buckets.
- Have a factory function to create the chooser given a date.
- The whole metadata chooser or the factory can be faked for testing.

pipeline/metadata/ip_metadata_chooser.py
ohnorobo: I think combining the different sub-pieces of the metadata classes here would be a mistake. 

Currently we have
file | lines
----|---
caida_ip_metadata.py | 370
test_caida_ip_metadata.py | 91
dbip.py | 70
maxmind.py | 116
mmdb_reader.py | 27
ip_metadata_chooser.py | 110

And in the future we plan to add more sub-metadata modules like AS size and possibly other whois information.

I think separating out
- how we get individual ip metadata from a specific source
from
- how we combine different ip metadata sources into a single addition to the row
is very important, and will only get more important in the future as we add more sources.

fortuna: A lot of that is boilerplate. You wouldn't lose any injection power.
But I'll leave that up to you. I do want, however, to have a factory function and the objects injected in the IPMetadataChooser constructor, since that's a lot clearer than having the two-stage initialization.

ohnorobo: Agreed about switching to the factory. I've added that.

I also figured out I can just subclass the fake off the real class instead of using an interface for both. That allows be to retain the type checking with less boilerplate.

pipeline/beam_tables.py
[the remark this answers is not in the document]
ohnorobo: done

pipeline/metadata/caida_ip_metadata.py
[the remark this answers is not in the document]
ohnorobo: switched to a structure with

    class CaidaIpMetadata():

and

    class FakeCaidaIpMetadata(CaidaIpMetadata):

instead of the interfaces

pipeline/metadata/ip_metadata_chooser.py
```diff
@@ -0,0 +1,104 @@
...
+    Args:
+      date: date to initialize caida metadata at
+      caida_ip_metadata_class: an IpMetadataInterface subclass (class, not instance)
+      caida_ip_metadata_bucket_folder: gcs folder with CAIDA ip metadata files
+      maxmind_class: an IpMetadataInterface subclass (class, not instance)
+      maxmind_bucket_folder: gcs folder with maxmind files
+      dbip_class: a DbipMetadata class (class, not instance)
+      dbip_bucket_folder: gcs folder with dbip files
+    """
+    self.caida = caida_ip_metadata_class(date, caida_ip_metadata_bucket_folder,
```
fortuna, line 39: Move these to the factory function and pass the caida, maxmind and dbip objects in the constructor, instead of class + bucket. No need for this class to know how they are created.

pipeline/metadata/ip_metadata_chooser.py
```diff
@@ -0,0 +1,104 @@
...
+      #    country) = maxmind_db.lookup(ip)
+      #   metadata_values['country'] = country
+    except KeyError as e:
+      logging.warning('KeyError: %s\n', e)
+      metadata_values = {}  # values are missing, but entry should still exist
+
+    return metadata_values
+
+
+class IpMetadataChooserFactory():
```
fortuna, line 79: This doesn't have to be a class. It could just be a function that takes a date and returns IpMetadataChooser.
Up to you, but I think the function makes it easier to create them

pipeline/metadata/ip_metadata_chooser.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+class FakeIpMetadataChooserFactory(IpMetadataChooserFactory):
+
+  # pylint: disable=super-init-not-called
+  def __init__(self) -> None:
+    pass
+
+  def make_chooser(self, date: datetime.date) -> IpMetadataChooser:
+    return IpMetadataChooser(date, caida_ip_metadata.FakeCaidaIpMetadata, "",
```
fortuna, line 102: The Fake classes don't need the bucket address, but you are passing it anyway.
By injecting the objects instead, you clean this issue.

fortuna (changes requested):

pipeline/metadata/ip_metadata_chooser.py
ohnorobo: I made it a class to make it easier to use the type in ScanDataBeamPipelineRunner.__init__

pipeline/metadata/ip_metadata_chooser.py
ohnorobo: done

pipeline/metadata/ip_metadata_chooser.py
ohnorobo: simplified these constructors to take nothing

pipeline/metadata/ip_metadata_chooser.py
```diff
@@ -0,0 +1,89 @@
...
+      metadata_values = {
+          'netblock': netblock,
+          'asn': asn,
+          'as_name': as_name,
+          'as_full_name': as_full_name,
+          'as_class': as_type,
+          'country': country,
+      }
+
+      (org, dbip_asn) = self.dbip.lookup(ip)
```
fortuna, line 39: It may be a little clearer now how you could simply pass a DBIP `mmdb_reader` in the constructor and not need the `DbipMetadata` library. But this looks good.

fortuna (approved):
This is so much cleaner now. Thanks for the PR!

pipeline/metadata/ip_metadata_chooser.py
ohnorobo: I'd like to do something smarter with DbipMetadata in the future (maybe use multiple tables, we'll see) So I'd like to keep the separation.

PR #52, opened 2021-08-20
Fix IP organizations

fortuna:
This fixes the issue in https://github.com/censoredplanet/jigsaw-cp-tracker/issues/20

The prod tables are updated.

I also added a filter on the data, since otherwise we need to read the outcome column for every dropdown on the dashboard.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -1,5 +1,9 @@
 # Development
 
+Install the dependencies with 
+
+    pip instal -r requirements.txt
+
 ## Testing Changes
 
 All tests and lint checks will be run on pull requests using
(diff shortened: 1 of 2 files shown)


docs/development.md
```diff
@@ -1,31 +1,35 @@
 # Development
 
+Install the dependencies with 
+
+    pip instal -r requirements.txt
```
ohnorobo, line 5: install

ohnorobo (approved):

docs/development.md
fortuna: Fixed

PR #53, opened 2021-08-23
switching country data to full names in the derived column

ohnorobo:
This gets us country names in the dashboard.

I tried to do something clever with keeping the country codes in the `_no_as` table and loading the names as a join, but it made table performance noticeably slower. This just replaces the codes in the country column with the full country names.

This does make the table slightly bigger
before: 1.81 TB
now: 1.87 TB

[test dashboard](https://datastudio.google.com/c/u/0/reporting/bc168cfd-13af-47c1-b6a9-e9ba7d9744d8/page/T33gB)

### DIFF
--- a/table/queries/merged_reduced_scans.sql
+++ b/table/queries/merged_reduced_scans.sql
@@ -116,7 +116,6 @@ CREATE TEMP FUNCTION ClassifyError(error STRING,
   END
 );
 
-
 CREATE OR REPLACE TABLE `firehook-censoredplanet.derived.merged_net_as`
 PARTITION BY date
 CLUSTER BY netblock, as_name, asn
@@ -136,7 +135,7 @@ AS (
 
 CREATE OR REPLACE TABLE `firehook-censoredplanet.derived.merged_reduced_scans_no_as`
 PARTITION BY date
-CLUSTER BY source, country, organization, domain
+CLUSTER BY source, country_name, organization, domain
 AS (
 WITH
   AllScans AS (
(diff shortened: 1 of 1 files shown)


table/queries/merged_reduced_scans.sql
```diff
@@ -194,14 +196,15 @@ WITH
...
-    country,
+    country_name as country,
     netblock,
     organization,
     controls_failed,
     CleanError(error) AS result,
     ClassifyError(error, "HTTPS", success, blockpage, page_signature) as outcome,
     count(1) AS count
   FROM `firehook-censoredplanet.base.https_scan`
+  JOIN `firehook-censoredplanet.metadata.country_names` ON country_code = country
```
fortuna, line 207: Use a `LEFT JOIN` instead to be clearer and more correct. If a country name is missing I want it to still show up.

table/queries/merged_reduced_scans.sql
```diff
@@ -194,14 +196,15 @@ WITH
...
   GROUP BY date, source, country, category, domain, netblock, organization, controls_failed, result, outcome
 )
 SELECT *
```
fortuna, line 210: Do the left join here instead to avoid the duplication.

I think you can do something like:
```sql
SELECT
  AllScans.* except (country),
  IFNULL(country_name, country) as country_name
FROM AllScans
LEFT JOIN `firehook-censoredplanet.metadata.country_names` ON country_code = country
```
Let's rename the field to `country_name` instead of `country` to avoid confusion, especially given `country` means different things in the pipeline. Ideally we would also change the pipeline to use `country_code` instead of `country`, but no need to do it now.

The IFNULL is so that we can see missing mapping.

fortuna (changes requested):

table/queries/merged_reduced_scans.sql
ohnorobo: done

table/queries/merged_reduced_scans.sql
ohnorobo: done

fortuna (approved):

PR #56, opened 2021-08-24
parameterize dataset names over the query

ohnorobo:
This makes it easier to write query results to a different table when developing without having to edit table names in the query.

The eventual goal here is to add running the query and testing the output table to the e2e test. But I wanted to get eyes on it before I wrote too much code.

I couldn't figure out a reasonable way to use `EXECUTE IMMEDIATE` without turning the query itself into a string.
https://stackoverflow.com/questions/59372469/create-table-using-a-variable-name-in-the-bigquery-ui But I'd like to find a less error-prone solution than the current basic string replace I have here. (While hopefully preserving sql syntax highlighting)

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -1,6 +1,6 @@
 # Development
 
-Install the dependencies with 
+Install the dependencies with
 
     pip install -r requirements.txt
 
@@ -61,6 +61,22 @@ If neither is specified the pipeline will automatically pick some reasonable
 dates to help avoid unintentially running test pipelines over large amounts of
 data. For appending pipelines this will be the latest week of data, and for
 full pipelines it will be the previous week of data.
+
+## Running queries
+
+After recreating the base tables you can rebuild any derived tables by running
+
+    python3 -m table.run_queries
+
+To test changes to the queries without overwriting the derived tables write to
+a personal dataset like
+
+    python3 -m table.run_queries --output=<username>
+
+If you want to read your own base tables (created from running a user pipeline
+above.) Then write the query as
+
+    python3 -m table.run_queries --input=<username> --output=<username>
 
 ## Access
 
--- a/table/run_queries.py
+++ b/table/run_queries.py
@@ -15,9 +15,10 @@
 
 Run as
 
-python3 tables/run_queries.py
+python3 -m table.run_queries
 """
 
+import argparse
 import glob
 from pprint import pprint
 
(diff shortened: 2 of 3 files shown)


table/run_queries.py
```diff
@@ -27,21 +28,48 @@
 
 client = cloud_bigquery.Client(project=firehook_resources.PROJECT_NAME)
 
+SOURCE_PLACEHOLDER = 'SOURCE'
```
fortuna, line 31: Can you make these placeholders more unique (e.g. `$BASE_DATASET$`)? I can see them possibly colliding if we don't pay attention.

docs/development.md
```diff
@@ -62,6 +62,22 @@ dates to help avoid unintentially running test pipelines over large amounts of
...
+## Running queries
+
+After recreating the base tables you can rebuild any derived tables by running
+
+    python3 -m table.run_queries
+
+To test changes to the queries without overwriting the derived tables write to
+a personal dataset like
+
+    python3 -m table.run_queries --sink=<username>
```
fortuna, line 74: I'd use  `input` and `output` instead of `source` and `sink`. I believe it's more common and clearer. Or perhaps we can be more explicit and say `base_dataset` and `derived_dataset`

table/run_queries.py
```diff
@@ -27,21 +28,48 @@
...
+SINK_PLACEHOLDER = 'SINK'
 
-def run_query(filepath: str) -> cloud_bigquery.table.RowIterator:
+DEFAULT_SOURCE_DATASET = 'base'
+DEFAULT_SINK_DATASET = 'derived'
+
+
+def _run_query(filepath: str, source: str,
```
fortuna, line 38: Let's use `base_dataset` and `derived_dataset` to be clearer

fortuna (changes requested):
I guess an alternative to this is to separate each query and specify the

docs/development.md
ohnorobo: done

table/run_queries.py
ohnorobo: done

table/run_queries.py
ohnorobo: switched to BASE_DATASET and DERIVED_DATASET. I tried to find a special character like $ that I could use. But everything I tried messed up standard sql syntax highlighting, which would make this file extremely annoying to edit going forward.

fortuna (approved):

PR #63, opened 2021-08-31
add file that defines metadata.country_names

ohnorobo:
I previously made the side-table firehook-censoredplanet.metadata.country_names manually

I wanted to check the file in here just to have it under source control. Unfortunately there doesn't seem to be a way to create it programmatically from the csv using DDL when we run the query. 

I also simplified some of the country names from the previous version. ("Iran, Islamic Republic of" -> just "Iran", etc) I also added a few missing ones. Those are currently updated in the table.

### DIFF
--- a/table/iso-country-codes.csv
+++ b/table/iso-country-codes.csv
@@ -1,0 +1,248 @@
+Afghanistan,AF
+Åland Islands,AX
+Albania,AL
+Algeria,DZ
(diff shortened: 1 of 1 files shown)


fortuna (approved):
You could potentially create a Github action that pushes this table to bigquery on every change, but we would need to figure it out.

PR #64, opened 2021-09-01
add versioning to merged table

ohnorobo:
Update the query to point to the `_v2` version of the table currently backing the dashboard.

Add a view which will act as a stable name for the table.


docs/merged_reduced_scans_table.md
```diff
@@ -1,7 +1,14 @@
...
 
-The table `firehook-censoredplanet.derived.merged_reduced_scans` contains filtered and pre-aggregated
-data to be used in the Censored Planet dashboard.
+The table `firehook-censoredplanet.derived.merged_reduced_scans_vN` (where N is a version number)
+contains filtered and pre-aggregated data to be used in the Censored Planet dashboard.
+
+This table is evolving and will change in backwards incompatible ways. When a backwards-incompatible
+change happens the version number will increment. Older version of the table will be deleted.
+
+The view `firehook-censoredplanet.derived.merged_reduced_scans` points to the latest version of
```
fortuna, line 9: Do you wanna name it `merged_reduced_scans_latest` to make it clearer what it is?
It may be confusing to tell the difference otherwise.

fortuna (approved):

docs/merged_reduced_scans_table.md
ohnorobo: I was hoping to to keep the old name so I don't have to go through and reconnect the data sources on all my side tables/old diagnostic queries.

fortuna: Ok, but you will still have to do that anyway, since the data change was a breaking change.

PR #65, opened 2021-09-01
streamline outcome names

ohnorobo:
Trying to make the mismatch error names clearer. Is there a clearer/more specific name we could use here than `mismatch` or `unexpected`?

- Echo, Discard, and HTTP/S (when we don't have more detail) all now use the same `content/mismatch` outcome. This will help the HTTP tab to have fewer different outcomes.
- HTTP/S (when more detail is available) use a longer string like `content/body_mismatch`. I think this helps make it more clear that `body_mismatch` results are a special case of general `unexpected` results.

Removed some Hyperquack v1/v2 distinctions in the documentation and tried to add more detail about when content is considered unexpected without getting too into the weeds.

Also tried to add a more representative examples for network/subnetwork.

### DIFF
--- a/docs/merged_reduced_scans_table.md
+++ b/docs/merged_reduced_scans_table.md
@@ -101,10 +101,10 @@ Mismatch Errors are used when the connection is successful, but the content rece
 | http.empty              | Received no content when HTTP content was expected |
 | http.truncated_response | The HTTP response content was unexpectedly truncated |
 |                         |
-| **Unexpected Content**  | The connection completed successfully, but the content returned didn't match the content expected for the domain. |
+| **Mismatched Content**  | The connection completed successfully, but the content returned didn't match the content expected for the domain. |
 |                         |
-| unexpected              | Received a different response from the one expected. </br> For Discard no response is expected and any response is a mismatch, </br> for Echo a mirrored response is expected and anything else is a mismatch. </br> For HTTP/S the expected response is determined by sending multiple control domains to the server and building an expected template. This response is returned when more detail about the exact part of the template mismatched (eg. status, body) is not available. |
-| unexpected_status       | The HTTP status code didn't match, eg. `403` instead of `200` |
-| unexpected_body         | The HTTP body didn't match, potentially a blockpage |
-| unexpected_tls          | An element of the TLS connection (certificate, cipher suite, or TLS version) didn't match |
+| mismatch                | Received a different response from the one expected. </br> For Discard no response is expected and any response is a mismatch, </br> for Echo a mirrored response is expected and anything else is a mismatch. </br> For HTTP/S the expected response is determined by sending multiple control domains to the server and building an expected template. This response is returned when more detail about the exact part of the template mismatched (eg. status, body) is not available. |
+| status_mismatch         | The HTTP status code didn't match, eg. `403` instead of `200` |
+| body_mismatch           | The HTTP body didn't match, potentially a blockpage |
+| tls_mismatch            | An element of the TLS connection (certificate, cipher suite, or TLS version) didn't match |
 | blockpage               | The response was unexpected and matched a [known blockpage]((https://github.com/censoredplanet/censoredplanet-analysis/blob/master/pipeline/metadata/data/blockpage_signatures.json)) |
--- a/docs/merged_reduced_scans_table.md
+++ b/docs/merged_reduced_scans_table.md
@@ -19,18 +19,18 @@ The columns `source` and `country_name` are always used for filtering, so they c
 
 Reduced Scans
 
-| Field Name | Type    | Contains |
-| ---------- | ------- | -------- |
-| date       | DATE    | Date that an individual measurement was taken |
-| source       | STRING    | What probe type the measurement came from ("HTTPS", "HTTP", "DISCARD", "ECHO") |
-| domain     | STRING  | The domain being tested (eg. `example.com`) or `CONTROL` for control measurements |
-| category   | STRING  | The [category](domain_categories.md) of the domain being tested, eg. `Social Networking`, `None` if unknown |
-| country_name    | STRING  | The country of the autonomous system, eg. `United States`  |
-| network        | STRING | The Autonomous System long name, eg. `Cloudflare, Inc.` |
-| subnetwork | STRING | The combinarion of the AS number and the IP organization, e.g. `AS6697 - Reliable Software, Ltd` |
-| outcome    | STRING  | An outcome classification, explained below. eg `read/timeout` |
-| count      | INTEGER | How many measurements fit the exact pattern of this row? |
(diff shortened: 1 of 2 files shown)


table/queries/merged_reduced_scans.sql
```diff
@@ -84,23 +84,23 @@ CREATE TEMP FUNCTION ClassifyError(error STRING,
...
-    WHEN (error = "Incorrect web response: status lines don't match") THEN "content/status_mismatch" # HTTP/S
-    WHEN (error = "Incorrect web response: bodies don't match") THEN "content/body_mismatch" # HTTP/S
-    WHEN (error = "Incorrect web response: certificates don't match") THEN "content/tls_mismatch" # HTTPS
-    WHEN (error = "Incorrect web response: cipher suites don't match") THEN "content/tls_mismatch" # HTTPS
-    WHEN (error = "Incorrect web response: TLS versions don't match") THEN "content/tls_mismatch" # HTTPS
+    WHEN (error = "Incorrect echo response") THEN "content/unexpected" # Echo
+    WHEN (error = "Received response") THEN "content/unexpected" # Discard
+    WHEN (error = "Incorrect web response: status lines don't match") THEN "content/unexpected_status" # HTTP/S
+    WHEN (error = "Incorrect web response: bodies don't match") THEN "content/unexpected_body" # HTTP/S
+    WHEN (error = "Incorrect web response: certificates don't match") THEN "content/unexpected_tls" # HTTPS
```
fortuna, line 93: I'd actually differentiate the TLS cases: `certificate_mismatch`, `cipher_mismatch`, `tls_version_mismatch`.

That may help us verify consistency and find MITM cases.

table/queries/merged_reduced_scans.sql
```diff
@@ -84,23 +84,23 @@ CREATE TEMP FUNCTION ClassifyError(error STRING,
...
-    WHEN (error = "Incorrect echo response") THEN "content/response_mismatch" # Echo
-    WHEN (error = "Received response") THEN "content/response_mismatch" # Discard
-    WHEN (error = "Incorrect web response: status lines don't match") THEN "content/status_mismatch" # HTTP/S
-    WHEN (error = "Incorrect web response: bodies don't match") THEN "content/body_mismatch" # HTTP/S
-    WHEN (error = "Incorrect web response: certificates don't match") THEN "content/tls_mismatch" # HTTPS
-    WHEN (error = "Incorrect web response: cipher suites don't match") THEN "content/tls_mismatch" # HTTPS
-    WHEN (error = "Incorrect web response: TLS versions don't match") THEN "content/tls_mismatch" # HTTPS
+    WHEN (error = "Incorrect echo response") THEN "content/unexpected" # Echo
+    WHEN (error = "Received response") THEN "content/unexpected" # Discard
+    WHEN (error = "Incorrect web response: status lines don't match") THEN "content/unexpected_status" # HTTP/S
```
fortuna, line 91: Not for this PR, but it would be great to output the status number here. I don't know if that's available.

docs/merged_reduced_scans_table.md
```diff
@@ -101,13 +101,10 @@ Mismatch Errors are used when the connection is successful, but the content rece
...
-| status_mismatch         | The HTTP status code didn't match, eg. `403` instead of `200` |
-| body_mismatch           | The HTTP body didn't match, potentially a blockpage |
-| tls_mismatch            | An element of the TLS connection (certificate, cipher suite, or TLS version) didn't match |
-| HyperQuack v1 and v2    |
-| template_mismatch       | Some element of the response did not match the expected template |
+| unexpected              | Received a different response from the one expected. </br> For Discard no response is expected and any response is a mismatch, </br> for Echo a mirrored response is expected and anything else is a mismatch. </br> For HTTP/S the expected response is determined by sending multiple control domains to the server and building an expected template. This response is returned when more detail about the exact part of the template mismatched (eg. status, body) is not available. |
+| unexpected_status       | The HTTP status code didn't match, eg. `403` instead of `200` |
```
fortuna, line 107: I don't think the new naming makes things clearer.

`unexpected_X` doesn't explain why `X` is unexpected. If I get status 404, is that an unexpected status?

I think the `X_mismatch` is better, because the `_mismatch` makes it clearer that you are comparing against something, the control.

Furthermore, it places the important thing, `X`, on the left, which can mitigate truncation issues in the legends.

fortuna: I see that all you are doing is renaming some outcomes and removing `template_mismatch`.
I think it's fine to still remove `template_mismatch`, but I'd replace it with the generic `response_ mismatch`, and keep the other names.

fortuna (changes requested):

table/queries/merged_reduced_scans.sql
ohnorobo: These three together are [>.1% of all https outcomes](https://docs.google.com/spreadsheets/d/1HisIsJWbsiWjp9sI8cU7xTpVao2L1oB4chiNKykclWM/edit?resourcekey=0-azZzA3FIPEODKRWnp-hpsA#gid=655449305) which is why I had combined them. I could be convinced splitting is a good idea, but I'd like to see an actual case in the dashboard where it helps analyse something. I don't think it's related to the change I'm trying to make in this PR.

docs/merged_reduced_scans_table.md
ohnorobo: Good point about the truncation.

I switched them back to `content/*_mismatch`, but turned `content/template_mismatch` and `content/response_mismatch` into just `content/mismatch`.

I think that helps accomplish my goal of making it clearer that `content/*_mismatch` outcomes are special cases of `content/mismatch` outcomes. 

If you still prefer `content/response_mismatch` let me know and I'll change it back.

table/queries/merged_reduced_scans.sql
ohnorobo: agreed

docs/merged_reduced_scans_table.md
fortuna: I think that's fine.

fortuna (approved):

table/queries/merged_reduced_scans.sql
fortuna: I agree it doesn't have to be part of this PR. I guess if we start seeing TLS issues we can add those cases, but I'd rather have more cases initially and simplify as we gain understanding. The low number of cases means it won't affect most usage, but it may help finding MITM when it actually happens.

Not high priority. It may be worth investigating some of the existing TLS errors at some point. I wonder if we can see the Kazakhstan MITM.

ohnorobo: The Kazakhstan MITM isn't in the main dataset unfortunately.

fortuna (approved):

PR #66, opened 2021-09-01
filter measurements without countries

ohnorobo:
Remove measurements which don't have a country.

If the country code is null it means we weren't able to get an AS/country for the IP from the CAIDA data. That's something we should fix in the pipeline, but it's meaningless to dashboard users and adds noise.

Currently we have 17mil measurements with null country codes (over all time) or .5% of the total data.

### DIFF
--- a/table/queries/merged_reduced_scans.sql
+++ b/table/queries/merged_reduced_scans.sql
@@ -169,6 +169,7 @@ SELECT
     END as unexpected_count
     FROM Grouped
     LEFT JOIN `firehook-censoredplanet.metadata.country_names` using (country_code)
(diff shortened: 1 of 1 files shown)


ramakrishnansr (approved):
Is it worth it to maybe try and grab some of these geolocations from other sources? But then without the AS, they may not be useful anyway.

ohnorobo:
> Is it worth it to maybe try and grab some of these geolocations from other sources? But then without the AS, they may not be useful anyway.

I think we can get both country and AS information from other sources like Maxmind or DBIP. How exactly to integrate multiple sources is a larger piece of work.

PR #71, opened 2021-09-16
require output argument to be specified when running manual queries

ohnorobo:
Require the `output` to be a specified argument when manually running the queries.

This will help avoid anyone overwriting the prod table by accident.

Tested:
> python3 -m table.run_queries
>
> usage: run_queries.py [-h] [--input INPUT] --output OUTPUT
> run_queries.py: error: the following arguments are required: --output


fortuna (approved):
I like that.

PR #72, opened 2021-09-17
treat known akamai responses as expected

ohnorobo:
Gave a different outcome to responses that are both from a known Akamai server header and send `x_on_this_server`. This is an attempt to start with the most conservative change we can make here that will still cause a large improvement in precision.

I went with giving the Akamai responses their own named outcome `expected/trusted_host` and treating them as expected (not just null) in the expected counts. This means they show up in the totals for the pivot table.

[test dashboard](https://datastudio.google.com/c/u/0/reporting/40dc493c-9974-415c-93ec-bab04de93b6d/page/T33gB) (I'm also experimenting in this dashboard with changing the outcome colors to match our sense of how unexpected something is. green=good, red=bad, yellow=maybe. Let me know how you like it.)

Example view of the UAE (surfaces blocking on AS5384 instead of being overwhelmed by Akamai errors)
Bcy5scVzsPhpooA

Example view of the US: (only shows the low level of non-`x-on-this-server` errors from Akamai
5697ZkE3JetihBw

### DIFF
--- a/table/queries/merged_reduced_scans.sql
+++ b/table/queries/merged_reduced_scans.sql
@@ -24,16 +24,20 @@ CREATE TEMP FUNCTION ClassifyError(error STRING,
                                    source STRING,
                                    template_match BOOL,
                                    blockpage_match BOOL,
-                                   blockpage_id STRING) AS (
+                                   blockpage_id STRING,
+                                   server_header STRING) AS (
   CASE
-    WHEN blockpage_match then CONCAT("content/blockpage:", blockpage_id)
+    WHEN blockpage_match THEN CONCAT("content/blockpage:", blockpage_id)
 
     # Content mismatch for hyperquack v2 which didn't write
     # content verification failures in the error field from 2021-04-26 to 2021-07-21
-    WHEN (NOT template_match AND (error is NULL OR error = "")) then "content/mismatch"
+    WHEN (NOT template_match AND (error IS NULL OR error = "")) THEN "content/mismatch"
+
+    # Trust 'x_on_this_server' responses when they come from Akamai servers.
+    WHEN blockpage_id = 'x_on_this_server' AND ((server_header = 'AkamaiGHost') OR (server_header = 'GHost')) THEN "content/known_mismatch"
 
     # Success
-    WHEN (error is NULL OR error = "") then "complete/success"
+    WHEN (error IS NULL OR error = "") THEN "complete/success"
 
     # System failures
     WHEN ENDS_WITH(error, "address already in use") THEN "setup/system_failure"
@@ -107,6 +111,13 @@ CREATE TEMP FUNCTION ClassifyError(error STRING,
   END
 );
 
+# Returns "abc" from the "Server: abc" header if it exists. Otherwise empty string.
+CREATE TEMP FUNCTION ExtractServerHeader(received_headers ARRAY<STRING>) AS (
+  IF(CONTAINS_SUBSTR(ARRAY_TO_STRING(received_headers, ','), "Server:"),
+     REGEXP_EXTRACT(CONCAT(ARRAY_TO_STRING(received_headers, ','),","), "Server: (.*?),"),
+     "")
+);
+
 # BASE_DATASET and DERIVED_DATASET are reserved dataset placeholder names
 # which will be replaced when running the query
 
@@ -126,31 +137,31 @@ OPTIONS (
 )
 AS (
 WITH AllScans AS (
-  SELECT * except (source), "DISCARD" AS source
+  SELECT * EXCEPT (source), "DISCARD" AS source
   FROM `firehook-censoredplanet.BASE_DATASET.discard_scan`
   UNION ALL
-  SELECT * except (source), "ECHO" AS source
+  SELECT * EXCEPT (source), "ECHO" AS source
   FROM `firehook-censoredplanet.BASE_DATASET.echo_scan`
   UNION ALL
-  SELECT * except (source), "HTTP" AS source
+  SELECT * EXCEPT (source), "HTTP" AS source
   FROM `firehook-censoredplanet.BASE_DATASET.http_scan`
   UNION ALL
-  SELECT * except (source), "HTTPS" AS source
+  SELECT * EXCEPT (source), "HTTPS" AS source
   FROM `firehook-censoredplanet.BASE_DATASET.https_scan`
 ), Grouped AS (
     SELECT
         date,
 
         source,
-        country as country_code,
-        as_full_name as network,
-        IF(is_control, "CONTROL", domain) as domain,
+        country AS country_code,
+        as_full_name AS network,
+        IF(is_control, "CONTROL", domain) AS domain,
 
-        ClassifyError(error, source, success, blockpage, page_signature) as outcome,
-        CONCAT("AS", asn, IF(organization is not null, CONCAT(" - ", organization), "")) as subnetwork,
+        ClassifyError(error, source, success, blockpage, page_signature, ExtractServerHeader(received_headers)) AS outcome,
+        CONCAT("AS", asn, IF(organization IS NOT NULL, CONCAT(" - ", organization), "")) AS subnetwork,
         category,
 
-        count(*) as count
+        COUNT(*) AS count
     FROM AllScans
     # Filter on controls_failed to potentially reduce the number of output rows (less dimensions to group by).
     WHERE NOT controls_failed
@@ -159,22 +170,24 @@ WITH AllScans AS (
     HAVING NOT STARTS_WITH(outcome, "setup/")
 )
 SELECT
-    Grouped.* except (country_code),
-    IFNULL(country_name, country_code) as country_name,
+    Grouped.* EXCEPT (country_code),
+    IFNULL(country_name, country_code) AS country_name,
     CASE
         WHEN (outcome = "complete/success") THEN 0
+        WHEN (outcome = "content/known_mismatch") THEN 0
         WHEN (STARTS_WITH(outcome, "dial/") OR STARTS_WITH(outcome, "setup/") OR ENDS_WITH(outcome, "/invalid")) THEN NULL
         WHEN (ENDS_WITH(outcome, "unknown")) THEN count / 2.0
         ELSE count
-    END as unexpected_count
+    END AS unexpected_count
     FROM Grouped
-    LEFT JOIN `firehook-censoredplanet.metadata.country_names` using (country_code)
+    LEFT JOIN `firehook-censoredplanet.metadata.country_names` USING (country_code)
     WHERE country_code IS NOT NULL
 );
 
 # Drop the temp function before creating the view
 # Since any temp functions in scope block view creation.
 DROP FUNCTION ClassifyError;
+DROP FUNCTION ExtractServerHeader;
 
 # This view is the stable name for the table above.
 # Rely on the table name firehook-censoredplanet.derived.merged_reduced_scans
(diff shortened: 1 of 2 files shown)


table/queries/merged_reduced_scans.sql
```diff
@@ -159,22 +170,24 @@ WITH AllScans AS (
     HAVING NOT STARTS_WITH(outcome, "setup/")
 )
 SELECT
-    Grouped.* except (country_code),
-    IFNULL(country_name, country_code) as country_name,
+    Grouped.* EXCEPT (country_code),
+    IFNULL(country_name, country_code) AS country_name,
     CASE
         WHEN (outcome = "complete/success") THEN 0
```
fortuna, line 176: I believe it's time to update how we output expected results.
It would be helpful to make it clearer what's expected.
In this case, we could have `expected/matches_control` and `expected/trusted_host:akamai`
The unexpected_count formula could use `STARTS_WITH('expected/')`

table/queries/merged_reduced_scans.sql
```diff
@@ -24,16 +24,20 @@ CREATE TEMP FUNCTION ClassifyError(error STRING,
...
+                                   blockpage_id STRING,
+                                   server_header STRING) AS (
   CASE
-    WHEN blockpage_match then CONCAT("content/blockpage:", blockpage_id)
+    WHEN blockpage_match THEN CONCAT("content/blockpage:", blockpage_id)
 
     # Content mismatch for hyperquack v2 which didn't write
     # content verification failures in the error field from 2021-04-26 to 2021-07-21
-    WHEN (NOT template_match AND (error is NULL OR error = "")) then "content/mismatch"
+    WHEN (NOT template_match AND (error IS NULL OR error = "")) THEN "content/mismatch"
```
fortuna, line 34: Won't this `content/mismatch` take precedence over the Akamai test?

table/queries/merged_reduced_scans.sql
```diff
@@ -107,6 +111,13 @@ CREATE TEMP FUNCTION ClassifyError(error STRING,
   END
 );
 
+# Returns "abc" from the "Server: abc" header if it exists. Otherwise empty string.
+CREATE TEMP FUNCTION ExtractServerHeader(received_headers ARRAY<STRING>) AS (
+  IF(CONTAINS_SUBSTR(ARRAY_TO_STRING(received_headers, ','), "Server:"),
```
fortuna, line 116: I believe there's no need to concatenate here. That may make the query significantly slower and harder to read.

You should be able to do something like 
```sql
select REGEXP_EXTRACT(header, "Server: (.*?)")
from UNNEST(received_headers) as header
where STARTS_WITH(header, "Server: ") LIMIT 1
```

Example: http://google3/jigsaw/firehook/metrics/update_intra_windowed_stats.sql;l=4;rcl=323798496

See https://cloud.google.com/bigquery/docs/reference/standard-sql/arrays#scanning_for_values_that_satisfy_a_condition

table/queries/merged_reduced_scans.sql
```diff
@@ -24,16 +24,20 @@ CREATE TEMP FUNCTION ClassifyError(error STRING,
...
-    WHEN blockpage_match then CONCAT("content/blockpage:", blockpage_id)
+    WHEN blockpage_match THEN CONCAT("content/blockpage:", blockpage_id)
 
     # Content mismatch for hyperquack v2 which didn't write
     # content verification failures in the error field from 2021-04-26 to 2021-07-21
-    WHEN (NOT template_match AND (error is NULL OR error = "")) then "content/mismatch"
+    WHEN (NOT template_match AND (error IS NULL OR error = "")) THEN "content/mismatch"
+
+    # Trust 'x_on_this_server' responses when they come from Akamai servers.
+    WHEN blockpage_id = 'x_on_this_server' AND ((server_header = 'AkamaiGHost') OR (server_header = 'GHost')) THEN "content/known_mismatch"
```
fortuna, line 37: What makes us trust the response is the fact that it came from Akamai. It wasn't injected.
That's true for all domains, hosted or not hosted on Akamai.

I believe it would be simpler to understand and explain if we just mark all tests that have the Akamai headers, and ignore the body pattern.

More importantly, we should prevent "expected" Akamai results from biasing the numbers when an AS happens to have Akamai, but it's not in the org id. Currently you will just remove the previously "unexpected" results.

We should also be more transparent. Let's output what the server is. The outcome could be `expected/trusted_host:akamai`

fortuna: One other thing that you don't need to address now: it's misleading to call the pattern id blockpage_id, when it isn't a block page. If the ids will include things that are not blockpages, we should rename it.

fortuna (changes requested):

table/queries/merged_reduced_scans.sql
fortuna: While at it, we might as well add Cloudflare. I think they use `Server: cloudflare`.
Another reason to always return the known host outcome is that it won't change depending on whether DDoS protection kicks in.

fortuna:
By the way, I do like the color idea. It will help visualize what's expected or not.

table/queries/merged_reduced_scans.sql
fortuna: Example of a Cloudflare page and its headers: https://explorer.ooni.org/measurement/20210817T073835Z_webconnectivity_KH_131178_n1_C08y4tgoO0sgboxv?input=http%3A%2F%2Flgbt.foundation%2F

ohnorobo: I'm specifically not trying to trust every server that returns an `AkamaiGHost` or `GHost` header. Those servers return a [bunch of other](https://datastudio.google.com/c/u/0/reporting/55e6ab82-8138-4ccc-a41f-66e6fb33d126/page/p_xd8o7a52mc?params=%7B%22df26%22:%22include%25EE%2580%25800%25EE%2580%2580IN%25EE%2580%2580HTTP%25EE%2580%2580HTTPS%22,%22df24%22:%22include%25EE%2580%25800%25EE%2580%2580IN%25EE%2580%2580AkamaiGHost%25EE%2580%2580GHost%22,%22df12%22:%22ORexclude%25EE%2580%25800%25EE%2580%2580IN%25EE%2580%2580content%252Fknown_response%253Ax_on_this_server%25EE%2580%2581include%25EE%2580%25803%25EE%2580%2580NU%25EE%2580%2582%22%7D) error patterns beyond just `x_on_this_server` including many unknown 3/400 errors. Many of those could be false positives but I haven't analysed them in any detail, and I wouldn't want to filter them out of the dashboard without looking more closely.

The point of this change is to do the most conservative thing (filtering out only the union of akamai server header and `x_on_this_server`) that we nevertheless know will make a large dashboard improvement.

fortuna: What I'm trying to get at is a simplified methodology that is easier to explain, understand and maintain.
It's a lot easier to say we are marking all Akamai responses as trusted_host:Akamai than saying we mark as that for some list of patterns that were arbitrarily selected and is not comprehensive. It's also a lot of work to vet all the patterns.

The assumption of my proposal is that censorship equipment will not try to impersonate other servers. That's not known to happen, and a mistake would treat sites as not blocked, so it's relatively safe to assume.

By marking all Akamai responses as expected we:

1. simplify the methodology. A simple methodology is important for people using the dashboard or trying to reproduce it. I appreciate the precaution to change less results, but I'd argue that a change that makes the methodology more complex is less conservative. It's more things to consider.
1. can filter out all of Akamai to remove bias, which is not possible if you are only marking some of the akamai results.
1. increase the coverage of errors that we mark as expected and the coverage is actually explainable as opposed to "some patterns we picked".
1. covers extra cases like the server successfully returning the actual site content, but not matching the control for some reason (e.g. regional content). You won't have patterns for those, other than the Server header or something else that says it's Akamai.

fortuna: As a side note, I'd prefer to not introduce any pattern that is based on the site content to the methodology, since that's really hard to scale and maintain.

Patterns for block pages are OK, since they are well known and obvious, and it's `O(censors)`, not `O(sites)`.
Patterns for hosting providers are ok too, since `num_hosting_providers` is much less than `num_sites`.

ramakrishnansr: First off, I agree with making the outcome be `expected/trusted_response:akamai` or `expected/trusted_response:pattern_id`, which makes it easier for users to connect with what we're matching to. Ideally, we should have all the patterns that we match against (be it response header or body) somewhere in this repository (or online) so that people can see what we're matching to. 

On whether we should mark all akamai responses as expected, I would say yes. Like I discussed, I think it's better to be conservative, and give users the default view of what we see (even if it is a censor, it's still an akamai page), and then if we get some knowledge of a pattern being incorrectly considered as expected, we can change it. But that should be the manual part, we should automatically mark these cases as expected. 

Also, I believe it is a really reasonable approach to match with the body patterns as well. There are plenty of cases where the server header is empty or contains different values, but the body still contains a very clear sign of a response from a hosting provider. We can find this by looking at how frequently a particular pattern appears. We shouldn't reduce the accuracy of the dashboard just because it is complicated to explain. I agree it may be a little hard to scale, but we've managed to have good coverage over 3 years, and the churn isn't as much because we consistently use the same IP addresses for measurements over time. 

Best approach to me seems: Have a public file of hosting provider responses which we consider as trusted, and match against all of them to conservatively mark as many expected responses as we can. This should be easy know for users to reproduce.

fortuna: I can see some patterns being useful, like the Cloudflare DDoS protection page. However, I'm curious to see cases where we can't classify looking at the headers alone. It's also safer to have patterns that are signature of specific hosting providers than generic patterns like `x_on_this_server`. However, this is a partial classification and it would mean biased numbers if we remove the CDN results that we know.
It's a tradeoff, and it will be interesting to explore more when we find such examples.

table/queries/merged_reduced_scans.sql
ohnorobo: It wasn't in my data, but it might in the time period where we're missing hyperquack v2 error output. Moved it above.

table/queries/merged_reduced_scans.sql
ohnorobo: Changed to 

`expected/match`
and
`expected/trusted_host:akamai`

table/queries/merged_reduced_scans.sql
ohnorobo: done

table/queries/merged_reduced_scans.sql
ohnorobo: Switched this to only match on the server pattern, not `x_on_this_server`.

I'm holding off on adding other hosts like cloudflare for now.

fortuna (approved):
Looks great. Thanks Sarah!

table/queries/merged_reduced_scans.sql
ramakrishnansr: I'm wondering why we're only marking Akamai and not the other obvious cases of expected responses? I agree that the majority of expected responses failing the template are Akamai, but there is a non-significant number of others as well that can be quite easily removed with the same approach.

ramakrishnansr: I'm worried that we bias the results towards other cloud providers now.

ohnorobo: Sorry, didn't see this comment before I merged. 

I wanted to just focus on akamai in this PR since we know the akamai precision increase is so good. Obviously we do want to filter out other expected responses, (I can already see cloudflare in the data) but deciding exactly what to filter/how to split that between the existing patterns and adding stuff to this query, requires some more thought.

PR #77, opened 2021-10-15
map Satellite to outcomes

avirkud:
- process `answers_err.json` from Satellite v1, which contains timeouts and other erroneous measurements
- add rcode field for Satellite v1 ('0' when no error, else '-1) 
- separate SQL query to create the reduced Satellite scan table
- rcode outcomes are from [here](https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml). See tracker/#23  for the remaining outcomes.


table/queries/merged_reduced_satellite_scans.sql
```diff
@@ -0,0 +1,142 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Classify all errors into a small set of enums
+#
+# Input is a nullable error string and rcode string from the raw data
+#
+# Output is a string of the format "stage/outcome"
+# Documentation of this enum is at
+# https://github.com/censoredplanet/censoredplanet-analysis/blob/master/docs/tables.md#outcome-classification
```
ohnorobo, line 21: Add a new .md page which documents these outcomes/this table, parallel to the documentation at https://github.com/censoredplanet/censoredplanet-analysis/blob/master/docs/merged_reduced_scans_table.md#outcome-classes . We can reference that instead of this more general link.

table/queries/merged_reduced_satellite_scans.sql
```diff
@@ -0,0 +1,142 @@
...
+        country AS country_code,
+        as_full_name AS network,
+        IF(is_control, "CONTROL", domain) AS domain,
+
+        SatelliteOutcome(received, rcode, error, controls_failed, anomaly) as outcome
+        CONCAT("AS", asn, IF(organization IS NOT NULL, CONCAT(" - ", organization), "")) AS subnetwork,
+        IFNULL(category, "Uncategorized") AS category,
+
+        COUNT(1) AS count
+    FROM `firehook-censoredplanet.BASE_DATASET.satellite_scan`
```
ohnorobo, line 109: once we deploy code with this query we'll need to be sure this base table doesn't get deleted while we're changing it, or the nightly pipeline could fail at the end.

PR #82, opened 2021-11-01
refactor satellite methods into their own file

ohnorobo:
Pure refactor, just moving code.

We probably want to to do a bit more refactoring to better isolate the way we're calling satellite methods in `run_beam_pipeline`. But I didn't want to make the logic change here since this is already big.


avirkud (commented):
I would consider moving `DateIpKey`, `_merge_metadata_with_rows`, etc. to a separate file within `metadata`, since they're also used during the Hyperquack pipelines for CAIDA metadata. 

`satellite.DateIpKey` sounds specific to Satellite

ohnorobo:
> I would consider moving `DateIpKey`, `_merge_metadata_with_rows`, etc. to a separate file within `metadata`, since they're also used during the Hyperquack pipelines for CAIDA metadata.
> 
> `satellite.DateIpKey` sounds specific to Satellite

Good idea. Split those out into the metadata.beam_metadata.py file.

avirkud (approved):

PR #88, opened 2021-11-03
change satv2.2 e2e test files to be more representative

ohnorobo:
Switched out some of the satellite v2.2 data for more interesting data.

- changed 6 of the `results.json` rows to be error rows
- changed all of the `resolvers.json` rows to match the vantage point ips from `results.json`
- changed 5 of the `blockpages.json` rows to be error rows

Currently the blockpage processing seems to be working, but the base table test is failing with 4 rows instead of 10.

I think I must be missing something about when satellite error rows get filtered out of the table. Is there a better criteria I can use to select `results.json` rows which will end up in the final table?


avirkud:
@ohnorobo this seems like a bug :) Any measurements that only have errors aren't being yielded.

I think we should only exclude measurements if the initial trial to the control domain fails (-> there are no trials for the test domain).

avirkud:
[branch for fix](https://github.com/censoredplanet/censoredplanet-analysis/compare/refactor-satellite-2...satellite-error-bug)

ohnorobo:
> branch for fix

Great. Could you make that a separate PR and add some unit testing?

PR #89, opened 2021-11-04
Move date partitioning for Satellite received IP tagging

avirkud:
Added a `received_tagging` parameter to satellite processing functions. This is set to False in `run_beam_pipeline` if we are writing to a production table.

Also moved the date partitioning for the v2.2 format change directly before the received IP steps.

[test job](https://console.cloud.google.com/dataflow/jobs/us-west2/2021-11-04_13_44_19-3200200413342898360?project=firehook-censoredplanet)

### DIFF
--- a/pipeline/beam_tables.py
+++ b/pipeline/beam_tables.py
@@ -581,12 +581,11 @@ class ScanDataBeamPipelineRunner():
       lines = _read_scan_text(p, new_filenames)
 
       if scan_type == satellite.SCAN_TYPE_SATELLITE:
-        received_tagging = True
         # For Satellite v1 - v2.1, the received IP tags (e.g. asnum) are in a
         # separate file from results.json. The steps for adding these tags are
-        # slow and we skip them in production.
-        if is_prod_table(table_name):
-          received_tagging = False
+        # slow so we skip them by setting `received_tagging` to False.
+        # TODO: If code is refactored, set to True to turn received IP tagging  back on.
+        received_tagging = False
         # PCollection[Row], PCollection[Row]
         satellite_rows, blockpage_rows = satellite.process_satellite_lines(
             lines, received_tagging)
--- a/pipeline/beam_tables.py
+++ b/pipeline/beam_tables.py
@@ -568,16 +568,9 @@ class ScanDataBeamPipelineRunner():
       lines = _read_scan_text(p, new_filenames)
 
       if scan_type == satellite.SCAN_TYPE_SATELLITE:
-        # For Satellite v1 - v2.1, the received IP tags (e.g. asnum) are in a
-        # separate file from results.json. Added the `received_tagging` param
-        # because the steps for adding these tags were slow in us-west2.
-        #
-        # TODO: The Satellite pipeline works with received tagging in
-        # us-central1, so this might not be necessary now.
-        received_tagging = True
         # PCollection[Row], PCollection[Row]
         satellite_rows, blockpage_rows = satellite.process_satellite_lines(
-            lines, received_tagging)
+            lines)
 
         # PCollection[Row]
         rows_with_metadata = self._add_metadata(satellite_rows)
--- a/pipeline/beam_tables.py
+++ b/pipeline/beam_tables.py
@@ -299,19 +299,6 @@ def get_table_name(dataset_name: str, scan_type: str,
   return f'{dataset_name}.{scan_type}_{base_table_name}'
 
 
-def is_prod_table(table_name: str) -> bool:
-  """Return true if the bigquery table is a production table.
-
-  Args:
-    table_name: table name with format 'dataset.table' like 'base.echo_scan'
-  Returns:
-    boolean
-  """
-  # Datasets should not contain '.' (following linux username convention)
-  dataset = table_name.split('.')[0]
-  return dataset == PROD_DATASET_NAME
-
-
 def _raise_exception_if_zero(num: int) -> None:
   if num == 0:
     raise Exception("Zero rows were created even though there were new files.")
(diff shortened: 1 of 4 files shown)


pipeline/beam_tables.py
```diff
@@ -568,9 +581,15 @@ def run_beam_pipeline(self, scan_type: str, incremental_load: bool,
       lines = _read_scan_text(p, new_filenames)
 
       if scan_type == satellite.SCAN_TYPE_SATELLITE:
+        received_tagging = True
+        # For Satellite v1 - v2.1, the received IP tags (e.g. asnum) are in a
+        # separate file from results.json. The steps for adding these tags are
```
ohnorobo, line 586: I don't think it's a good idea to have seperate behavior for dev and prod here. We're just going to end up pushing unnoticed bugs what didn't show up in dev.

If we think these tags aren't useful enough to be worth the cost and expense in prod then we shouldn't add them in dev either. It sounds like the actual relevant distinction is that we want to avoid adding them for v2.2 data, but that code should all be able to live in satellite.py

Since we have to decide the structure of the pipeline ahead of runtime this may require doing something like adding a partitioning step into pre2.2 rows and post-2.2 rows, then only running the tagging steps on the pre-2.2 rows. That means when we run nightly jobs over only post-2.2 data the tagging steps will appear in the pipeline structure, but 0 rows will ever flow through it.

ohnorobo (changes requested):

ohnorobo (changes requested):
(github ate my comment, sorry if you get two version of this)

I don't think it's a good idea two have different behavior for dev and prod here. All that's going to do is introduce subtle bugs in the future.

It also seems like dev vs prod isn't even the distinction we want. The point is that we don't want to add tags for rows from v2.2 data.

Since the structure of the pipeline all has to be determined ahead of time and independently of which data we're running on I think the way to do this is to partition the data into pre-v2.2 and post-v2.2 sections, then only run the tagging on the pre-v2.2 sections. This means the tagging steps will exist even when we run post-v2.2 nightly pipelines, but 0 rows should run through those steps.

avirkud:
> I don't think it's a good idea to have different behavior for dev and prod here. All that's going to do is introduce subtle bugs in the future.
> 
Yeah, makes sense.

> It also seems like dev vs prod isn't even the distinction we want. The point is that we don't want to add tags for rows from v2.2 data.
> 
> Since the structure of the pipeline all has to be determined ahead of time and independently of which data we're running on I think the way to do this is to partition the data into pre-v2.2 and post-v2.2 sections, then only run the tagging on the pre-v2.2 sections. This means the tagging steps will exist even when we run post-v2.2 nightly pipelines, but 0 rows should run through those steps.


We use similar partitioning logic so post-v2.2 rows don't go through the received tagging steps. The issue I had with including received tagging in prod is that the steps are very slow for the v1-2.1 data. 

The test jobs typically took 2+ hours for a single scan, so I would expect a backfill to take several days at minimum. Currently we don't have any visualizations based on the received IP tags, so I don't know if it's worth the time whenever we need to run a backfill.

To keep behavior consistent between dev and prod, I think I would prefer to remove the received tagging steps from dev as well. We could add them back to both if we are able to refactor the steps. Let me know if what you think about this

ohnorobo:
Yes in that case I'd be down to just remove the tagging. (Is this only removing it in v2.1 and v2.2 then?) Maybe leave a note in the code on why it was removed and where to re-add it in the future if we can make it faster.

ohnorobo:
I think you have a weird merge going on. I'm seeing some changes from https://github.com/censoredplanet/censoredplanet-analysis/pull/91 showing up here. Could you try flattening all your changes down into a single commit?

avirkud:
I rebased incorrectly, which caused the duplicated commits. I think that's fixed now?

There was a bug in unflatten for the v2.2 data that was causing the [test jobs](https://console.cloud.google.com/dataflow/jobs/us-central1/2021-11-11_08_52_11-9773590606068567070?project=firehook-censoredplanet) to fail during post-processing. 

This now works with the e2e tests, just started a test job since your test backfill finished.

pipeline/beam_tables.py
```diff
@@ -568,9 +568,16 @@ def run_beam_pipeline(self, scan_type: str, incremental_load: bool,
       lines = _read_scan_text(p, new_filenames)
 
       if scan_type == satellite.SCAN_TYPE_SATELLITE:
+        # For Satellite v1 - v2.1, the received IP tags (e.g. asnum) are in a
+        # separate file from results.json. Added the `received_tagging` param
+        # because the steps for adding these tags were slow in us-west2.
+        #
+        # TODO: The Satellite pipeline works with received tagging in
+        # us-central1, so this might not be necessary now.
+        received_tagging = True
```
ohnorobo, line 577: Since this is currently always set to True I don't think we need to pass it through. 

Since `received_tagging = True` is working I don't think we want to run pipelines with it turned off. I would remove the places where this is plumbed through, and remove the branches where `received_tagging = False` (We also don't have testing for those branches, which makes me nervous.)

That will leave us with one common behavior: always tagging received IPs
And one major change in this PR: using the existing tag data for v2.2 rows instead of caida.

ohnorobo (changes requested):

pipeline/metadata/satellite.py
```diff
@@ -611,6 +621,7 @@ def process_satellite_lines(
 
   # PCollection[Row]
   tagged_satellite = process_satellite_with_tags(lines, tags)
+  # Post processing steps require the received IP tags.
```
ohnorobo, line 624: You can remove this comment now

ohnorobo (approved):

avirkud:
[Test backfill](https://console.cloud.google.com/dataflow/jobs/us-central1/2021-11-15_07_38_13-10760576185535988274?project=firehook-censoredplanet) succeeded!

PR #94, opened 2021-11-11
make filenames a pcollection instead of individual steps

ohnorobo:
This massively cuts down on the amount of time it takes to start up jobs with hundreds/thousands of input files.

This is only recently simple because `ReadAllFromText` added the `with_filenames` option https://github.com/apache/beam/pull/15126 / https://issues.apache.org/jira/browse/BEAM-12665

Also fix `test_read_scan_text`, which was erroneously passing but not actually checking anything before.

PipelineManualE2eTest passed.

I'm going to hold off on merging this until I get a chance to run a full satellite backfill with it. (To make sure it doesn't cause fusion.) Currently I'm waiting so i don't interfere with a job @avirkud is running.

### DIFF
--- a/pipeline/beam_tables.py
+++ b/pipeline/beam_tables.py
@@ -194,24 +194,13 @@ def _read_scan_text(
     A PCollection[Tuple[filename, line]] of all the lines in the files keyed
     by filename
   """
-  # List[PCollection[Tuple[filename,line]]]
-  line_pcollections_per_file: List[beam.PCollection[Tuple[str, str]]] = []
+  # PCollection[filename]
+  pfilenames = (p | beam.Create(filenames))
 
-  for filename in filenames:
-    # PCollection[line]
-    lines = p | 'read file ' + filename >> beam.io.ReadFromText(filename)
-    step_name = 'annotate filename ' + filename
-
-    # PCollection[Tuple[filename,line]]
-    lines_with_filenames = (
-        lines | step_name >> beam.Map(_make_tuple, filename).with_output_types(
-            Tuple[str, str]))
-    line_pcollections_per_file.append(lines_with_filenames)
-
-  # PCollection[Tuple[filename,line]]
+  # PCollection[Tuple(filename, line)]
   lines = (
-      tuple(line_pcollections_per_file) | 'flatten lines' >>
-      beam.Flatten(pipeline=p).with_output_types(Tuple[str, str]))
+      pfilenames | "read files" >> beam.io.ReadAllFromText(with_filename=True))
+
   return lines
 
 
--- a/pipeline/test_beam_tables.py
+++ b/pipeline/test_beam_tables.py
@@ -117,15 +117,22 @@ class PipelineMainTest(unittest.TestCase):
 
   def test_read_scan_text(self) -> None:  # pylint: disable=no-self-use
     """Test reading lines from compressed and uncompressed files"""
-    p = TestPipeline()
-    pipeline = beam_tables._read_scan_text(
-        p, ['pipeline/test_results_1.json', 'pipeline/test_results_2.json.gz'])
+    with TestPipeline() as p:
+      lines = beam_tables._read_scan_text(
+          p,
+          ['pipeline/test_results_1.json', 'pipeline/test_results_2.json.gz'])
 
-    beam_test_util.assert_that(
-        pipeline,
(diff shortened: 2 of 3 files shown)


pipeline/beam_tables.py
```diff
@@ -194,24 +194,13 @@ def _read_scan_text(
     A PCollection[Tuple[filename, line]] of all the lines in the files keyed
     by filename
   """
-  # List[PCollection[Tuple[filename,line]]]
-  line_pcollections_per_file: List[beam.PCollection[Tuple[str, str]]] = []
+  # PCollection[filename]
```
fortuna, line 197: I don't think you need `_read_scan_text`. The indirection hurts readability. I'd just inline the code in the caller.

pipeline/beam_tables.py
```diff
@@ -194,24 +194,13 @@ def _read_scan_text(
...
+  pfilenames = (p | beam.Create(filenames))
```
fortuna, line 198: Not for this PR, but consider creating the file list in the pipeline too.

pipeline/test_beam_tables.py
```diff
@@ -117,15 +117,22 @@ def test_get_full_table_name(self) -> None:
...
-    p = TestPipeline()
-    pipeline = beam_tables._read_scan_text(
-        p, ['pipeline/test_results_1.json', 'pipeline/test_results_2.json.gz'])
-
-    beam_test_util.assert_that(
-        pipeline,
-        beam_test_util.equal_to([
-            'test line 1.1', 'test line 1.2', 'test line 2.1', 'test line 2.2'
-        ]))
+    with TestPipeline() as p:
```
fortuna, line 120: You can delete this test if you delete `_read_scan_text`.

fortuna (approved):

pipeline/test_beam_tables.py
ohnorobo: I think this is a valuable thing to be testing. (Both that we can read files correctly and that we can read both zipped and unzipped files transparently.) We're not testing this functionality anywhere else in the unit tests, only the e2e tests which are substantially slower.

pipeline/beam_tables.py
ohnorobo: I'd like to keep this seperate for the purposes of keeping the unit test below. If it's inlined in the caller then we can't test this functionality independently.

pipeline/beam_tables.py
ohnorobo: I think there are some speedups I can make to creating the file list. But I'm mostly not sure what creating it in the pipeline would buy us. It's not like enumerating the list is any faster on a worker machine than it is for me during pipeline setup.

One major downside to doing that is we'll lose the ability to abort creating the pipeline in the first place when there's no new data like we do [here](https://github.com/censoredplanet/censoredplanet-analysis/blob/761f614833fe8f64d771e5a70f6d088dc35285d1/pipeline/beam_tables.py#L563). We'll also lose the ability to debug the content of the file list at startup time, since it will only be available inside the pipeline.

pipeline/beam_tables.py
ohnorobo: Generally anything that gets inlined into `run_beam_pipeline` becomes untestable outside of the e2e tests since `run_beam_pipeline` writes to bigquery.

pipeline/beam_tables.py
fortuna: Listing files should be a lot faster in the worker, because it's running from inside Cloud, ideally in the same region. 

You could also parallelize the different sources.

The abort is not needed to implement: if there are no files, you won't output anything and the pipeline will end.

If you'd like to debug the list of processed files, you can save that as a side output. I did that in the past to find broken OONI files (in that case it was a list of broken files).

By the way, there are other optimization opportunities. You are currently matching against the entire dataset, even for a single date. That's bad, as it's not proportional to the work that needs to be done.

Instead, it's possible to list directories in order starting with a key. Then you can just list the files in the range you need. You'd need the GCS Client. This is what I do for OONI (see StartAfter): https://github.com/Jigsaw-Code/net-analysis/blob/2b36fe89c3305f4d1c93406725a7f46a74f246f9/netanalysis/ooni/data/ooni_client.py#L91

In any case, not for this PR.

fortuna (approved):

PR #95, opened 2021-11-12
simplify filename matching

ohnorobo:
Streamline how we're enumerating which files to load for dataflow jobs.

This saves us from doing 22 gcs enumerations for each satellite job 😅 (one for all 11 [satellite filetypes](https://github.com/censoredplanet/censoredplanet-analysis/blob/761f614833fe8f64d771e5a70f6d088dc35285d1/pipeline/metadata/satellite.py#L69) multiplied by two for both zipped and unzipped files) and two for all hyperquack jobs, to just a single `gcs.match` per job.

### DIFF
--- a/pipeline/beam_tables.py
+++ b/pipeline/beam_tables.py
@@ -17,6 +17,7 @@ from __future__ import absolute_import
 
 import datetime
 import logging
+import pathlib
 import re
 from typing import Optional, Tuple, Dict, List, Any, Iterator, Iterable
 
@@ -240,6 +241,23 @@ def _between_dates(filename: str,
   return True
 
 
+def _filename_matches(filename: str, allowed_filenames: List[str]) -> bool:
+  """Find if a filename matches a list of filenames.
+
+  Args:
+    filename, zipped or unzipped ex: results.js, results.json.gz
+    allowed_filenames: List of filenames, all unzipped
+      ex: [resolvers.json, blockpages.json]
+
+  Returns:
+    boolean, whether the filename matches one of the list.
+    Zipped matches to unzipped names count.
+  """
+  if '.gz' in pathlib.PurePosixPath(filename).suffixes:
+    filename = pathlib.PurePosixPath(filename).stem
+  return filename in allowed_filenames
+
+
 def _get_partition_params(scan_type: str = None) -> Dict[str, Any]:
   """Returns additional partitioning params to pass with the bigquery load.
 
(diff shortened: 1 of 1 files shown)


pipeline/beam_tables.py
```diff
@@ -377,29 +395,21 @@ def _data_to_load(self,
...
-      zipped_metadata = [
-          m.metadata_list for m in gcs.match([zipped_regex.format(file)])
-      ][0]
-      unzipped_metadata = [
-          m.metadata_list for m in gcs.match([unzipped_regex.format(file)])
-      ][0]
-      file_metadata += zipped_metadata + unzipped_metadata
+    # Filepath like `gs://firehook-scans/echo/**/*'
+    files_regex = f'{self.bucket}{scan_type}/**/*'
+    file_metadata = [m.metadata_list for m in gcs.match([files_regex])][0]
```
fortuna, line 400: Nice change. gcs.match is expensive!

In the other PR I suggest a solution without match. You can use [google.cloud.storage.client.Client.list_blobs](https://googleapis.dev/python/storage/latest/client.html#google.cloud.storage.client.Client.list_blobs) with a start offset. You can stop listing when the end date is reached, or specify the end offset. Also, it returns an iterator, so you can start outputting filenames to be processed right away *before* the full listing is completed. The parallelism is significant.

Furthermore, the listing work is proportional to the work that needs to be done. That's the approach I use for the OONI data, and is a lot more efficient.

fortuna (approved):

pipeline/beam_tables.py
ohnorobo: (Also partially responding to your comment in https://github.com/censoredplanet/censoredplanet-analysis/pull/94)

It's not terribly expensive. Currently the slowest version of this call we make is
`gcs.match('gs://firehook-scans/satellite/**/*')`
which matches 4310 files

That's basically equivalent to running 
`gsutil ls gs://firehook-scans/satellite/**/*`

If I time that call I get 
`1.45s user 0.21s system 85% cpu 1.933 total`

Which is nothing on a pipeline that's going to take several minutes to get loaded into dataflow and probably several hours to run. (Still nice not to have to do it 22 times though.)

How long does it take you to enumerate the OONI files?

fortuna: OONI was way worse because we had subdirectories for each hour of each day, and then at least a file per country and test type (Example filename: `raw/20210508/04/TR/webconnectivity/2021050804_TR_webconnectivity.n0.0.jsonl.gz`). Listing could take a few minutes.

This single date took 14.5s:
```
time aws --no-sign-request s3 ls s3://ooni-data-eu-fra/raw/20210526/ --recursive
```

PR #104, opened 2021-11-22
strip .gz from satellite files

ohnorobo:
This is a fix for an error I introduced in https://github.com/censoredplanet/censoredplanet-analysis/pull/101

Runtime error visible in https://console.cloud.google.com/errors/CJzcht_zxtaPTg?project=firehook-censoredplanet&time=P30D

### DIFF
--- a/pipeline/metadata/flatten_satellite.py
+++ b/pipeline/metadata/flatten_satellite.py
@@ -191,72 +191,69 @@ class SatelliteFlattener():
     self.blockpage_matcher = blockpage_matcher
     self.category_matcher = category_matcher
 
-  def process_satellite(self, filename: str, scan: Any,
+  def process_satellite(self, filepath: str, scan: Any,
                         random_measurement_id: str) -> Iterator[Row]:
     """Process a line of Satellite data.
 
     Args:
-      filename: a filepath string
+      filepath: a filepath string
       scan: a loaded json object containing the parsed content of the line
       random_measurement_id: a hex id identifying this individual measurement
 
     Yields:
       Rows
     """
-    date = re.findall(r'\d\d\d\d-\d\d-\d\d', filename)[0]
-    if pathlib.PurePosixPath(filename).name == SATELLITE_BLOCKPAGES_FILE:
-      yield from self._process_satellite_blockpages(scan, filename)
+    date = re.findall(r'\d\d\d\d-\d\d-\d\d', filepath)[0]
+
+    filename = pathlib.PurePosixPath(filepath).name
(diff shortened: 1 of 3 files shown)


pipeline/metadata/flatten_satellite.py
```diff
@@ -191,72 +191,69 @@ def __init__(self, blockpage_matcher: BlockpageMatcher,
...
+
+    filename = pathlib.PurePosixPath(filepath).name
+    if '.gz' in pathlib.PurePosixPath(filename).suffixes:
+      filename = pathlib.PurePosixPath(filename).stem
+
+    if filename == SATELLITE_BLOCKPAGES_FILE:
+      yield from self._process_satellite_blockpages(scan, filepath)
     elif datetime.date.fromisoformat(date) < SATELLITE_V2_1_START_DATE:
-      yield from self._process_satellite_v1(date, scan, filename,
+      yield from self._process_satellite_v1(date, scan, filepath,
```
avirkud, line 215: I think you could pass `filename` here. Then you don't need to duplicate the three lines in `_process_satellite_v1` (240-242).

pipeline/metadata/flatten_satellite.py
```diff
@@ -191,72 +191,69 @@ def __init__(self, blockpage_matcher: BlockpageMatcher,
...
-    if pathlib.PurePosixPath(filename).name == SATELLITE_BLOCKPAGES_FILE:
-      yield from self._process_satellite_blockpages(scan, filename)
+    date = re.findall(r'\d\d\d\d-\d\d-\d\d', filepath)[0]
+
+    filename = pathlib.PurePosixPath(filepath).name
+    if '.gz' in pathlib.PurePosixPath(filename).suffixes:
+      filename = pathlib.PurePosixPath(filename).stem
+
+    if filename == SATELLITE_BLOCKPAGES_FILE:
+      yield from self._process_satellite_blockpages(scan, filepath)
```
avirkud, line 213: The docstring for `_process_satellite_blockpages` needs to be updated (also the parameter name -> filepath?).

avirkud (changes requested):

pipeline/metadata/flatten_satellite.py
ohnorobo: I'd rather keep passing the filepath and do the strip twice. 

I think we're going to want the filepath in the future. (For example to calculate a `source` field for the satellite table.)

pipeline/metadata/flatten_satellite.py
ohnorobo: updated

avirkud (approved):

PR #109, opened 2021-12-03
Separate out hyperquack specific schema columns

ohnorobo:
This separates out all the columns that are never filled in the satellite tables. (All yellow columns [here](https://docs.google.com/document/d/1BMohMUiqJzYnTR8fdEqb1Ang5r1oW_aEHdsxQraQPR8/edit?hl=en#)) 
Now they will only exist in the hyperquack data.

columns are:
- blockpage
- page_signature
- stateful_block
- received_status
- received_body
- received_headers
- received_tls_version
- received_tls_cipher_suite
- received_tls_cert

### DIFF
--- a/pipeline/beam_tables.py
+++ b/pipeline/beam_tables.py
@@ -31,6 +31,7 @@ from google.cloud import bigquery as cloud_bigquery  # type: ignore
 from pipeline.metadata.beam_metadata import DateIpKey, IP_METADATA_PCOLLECTION_NAME, ROWS_PCOLLECION_NAME, make_date_ip_key, merge_metadata_with_rows
 from pipeline.metadata.flatten import Row
 from pipeline.metadata import flatten_base
+from pipeline.metadata import flatten_hyperquack
 from pipeline.metadata import flatten_satellite
 from pipeline.metadata import flatten
 from pipeline.metadata import satellite
@@ -42,7 +43,7 @@ BASE_TABLE_NAME = 'scan'
 PROD_DATASET_NAME = 'base'
 
 # key: (type, mode)
-SCAN_BIGQUERY_SCHEMA = {
+BASE_BIGQUERY_SCHEMA = {
     # Columns from Censored Planet data
     'domain': ('string', 'nullable'),
     'category': ('string', 'nullable'),
@@ -53,23 +54,10 @@ SCAN_BIGQUERY_SCHEMA = {
     'error': ('string', 'nullable'),
     'anomaly': ('boolean', 'nullable'),
     'success': ('boolean', 'nullable'),
-    'stateful_block': ('boolean', 'nullable'),
     'is_control': ('boolean', 'nullable'),
     'controls_failed': ('boolean', 'nullable'),
     'measurement_id': ('string', 'nullable'),
     'source': ('string', 'nullable'),
-    'blockpage': ('boolean', 'nullable'),
-    'page_signature': ('string', 'nullable'),
-
-    # Column filled in all tables
-    'received_status': ('string', 'nullable'),
-    # Columns filled only in HTTP/HTTPS tables
-    'received_body': ('string', 'nullable'),
-    'received_headers': ('string', 'repeated'),
-    # Columns filled only in HTTPS tables
-    'received_tls_version': ('integer', 'nullable'),
-    'received_tls_cipher_suite': ('integer', 'nullable'),
-    'received_tls_cert': ('string', 'nullable'),
 
     # Columns added from CAIDA data
     'netblock': ('string', 'nullable'),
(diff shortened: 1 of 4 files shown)


pipeline/beam_tables.py
```diff
@@ -113,16 +101,24 @@ def _get_bigquery_schema(scan_type: str) -> Dict[str, Any]:
...
   """
   if scan_type == satellite.SCAN_TYPE_BLOCKPAGE:
     return satellite.BLOCKPAGE_BIGQUERY_SCHEMA
   if scan_type == satellite.SCAN_TYPE_SATELLITE:
-    full_schema: Dict[str, Any] = {}
-    full_schema.update(SCAN_BIGQUERY_SCHEMA)
-    full_schema.update(satellite.SATELLITE_BIGQUERY_SCHEMA)
-    return full_schema
-  return SCAN_BIGQUERY_SCHEMA
+    return add_schemas(BASE_BIGQUERY_SCHEMA,
```
fortuna, line 109: Instead of dynamically creating the schema, define them as constants. Example

```py
SATELLITE_BIGQUERY_SCHEMA = BASE_BIGQUERY_SCHEMA.copy().update(...)
```

Mutations add complexity and currently the HYPERQUACK/SATELLITE_SCHEMA are not a full schema, so the variable name is essentially lying, which hurts readability.

fortuna (changes requested):

pipeline/beam_tables.py
ohnorobo: Renamed the partial schemas as `ADDITIONAL_HYPERQUACK_BIGQUERY_SCHEMA`/`ADDITIONAL_SATELLITE_BIGQUERY_SCHEMA` and defined the full schemas as constants in `beam_tables.py`. I kept the `_add_schemas` function instead of doing an inline `.copy().update(...)` though to improve readability.

pipeline/test_beam_tables.py
```diff
@@ -24,26 +24,36 @@
...
+        list(flatten_hyperquack.ADDITIONAL_HYPERQUACK_BIGQUERY_SCHEMA.keys()))
+    self.assertListEqual(
+        list(echo_schema.keys()), all_hyperquack_top_level_columns)
 
+  def test_get_bigquery_schema_satellite(self) -> None:
     satellite_schema = beam_tables._get_bigquery_schema('satellite')
     all_satellite_top_level_columns = (
-        list(beam_tables.SCAN_BIGQUERY_SCHEMA.keys()) +
-        list(satellite.SATELLITE_BIGQUERY_SCHEMA.keys()))
+        list(beam_tables.BASE_BIGQUERY_SCHEMA.keys()) +
```
fortuna, line 51: I'd rather see here something like `HYPERQUACK_BIGQUERY_SCHEMA.keys()`.

It's a lot more readable. You are leaking how the schema is built, but the reader shouldn't have to care.

pipeline/beam_tables.py
```diff
@@ -86,23 +92,22 @@
...
+  full_schema.update(schema_a)
+  full_schema.update(schema_b)
+  return full_schema
 
-# Data files for the non-Satellite pipelines
-SCAN_FILES = ['results.json']
 
-# An empty json file is 0 bytes when unzipped, but 33 bytes when zipped
-EMPTY_GZIPPED_FILE_SIZE = 33
+HYPERQUACK_BIGQUERY_SCHEMA = _add_schemas(
```
fortuna, line 105: Why is this and ADDITIONAL_HYPERQUACK_BIGQUERY_SCHEMA in different files?
Define only the former. The later could be inlined in the _add_schemas call.

pipeline/metadata/flatten_hyperquack.py
```diff
@@ -14,6 +14,24 @@
 # echo/discard domain and url content
 SENT_PATTERN = "GET (.*) HTTP/1.1\r\nHost: (.*)\r\n"
 
+# Additional bigquery fields for the satellite data
+# These are in addition to the fields in beam_tables.BASE_BIGQUERY_SCHEMA
+ADDITIONAL_HYPERQUACK_BIGQUERY_SCHEMA = {
```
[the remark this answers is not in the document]
ohnorobo: I think having a named variable is more readable than inlining it, but okay.

pipeline/test_beam_tables.py
ohnorobo: done

pipeline/beam_tables.py
ohnorobo: Moved both schemas back to the beam tables file and inlined them.

pipeline/metadata/flatten_hyperquack.py
```diff
@@ -14,6 +14,24 @@
 # echo/discard domain and url content
 SENT_PATTERN = "GET (.*) HTTP/1.1\r\nHost: (.*)\r\n"
 
+# Additional bigquery fields for the satellite data
+# These are in addition to the fields in beam_tables.BASE_BIGQUERY_SCHEMA
```
fortuna, line 18: It seems this is unused code now?

pipeline/metadata/satellite.py
```diff
@@ -18,8 +18,8 @@
 from pipeline.metadata import flatten
 
 # Additional bigquery fields for the satellite data
-# These are in addition to the fields in beam_tables.SCAN_BIGQUERY_SCHEMA
-SATELLITE_BIGQUERY_SCHEMA = {
+# These are in addition to the fields in beam_tables.BASE_BIGQUERY_SCHEMA
+ADDITIONAL_SATELLITE_BIGQUERY_SCHEMA = {
```
fortuna, line 22: Delete?

pipeline/metadata/flatten_hyperquack.py
ohnorobo: Oh sorry, bad commit. Fixed.

pipeline/metadata/satellite.py
ohnorobo: fixed

fortuna (approved):

PR #110, opened 2021-12-06
remove global INTERFERENCE_IPDOMAIN which cannot work

ohnorobo:
The current code is using the global `INTERFERENCE_IPDOMAIN` to pass information between computation in `flatten_satellite.py` and `satellite.py`. Although this can be mocked out to work in unit tests it cannot possibly work when running a distributed pipeline since different workers won't have the same `INTERFERENCE_IPDOMAIN`.

Querying the actual data 'domain_below_threshold' never shows up in `exclude_reason`. So this is definitely not working.

To add this functionality back we'll have to figure out a way to calculate it (probably via a group) and pass the value in the pipeline.


ohnorobo:
Git is being weird and auto-closing my PRS for some reason.

PR #114, opened 2022-01-14
Fix bug setting anomaly to False

avirkud:
- updated the partitioning between test and control resolvers to use the `is_control_ip` field. This should now work correctly for v2.2, where the control measurements are not clearly distinguished in the raw data.
- removed the check where we set anomaly to False for cases of `untagged_controls`. This was causing the abnormally low percentages of anomalies.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -30,6 +30,6 @@ disable=abstract-method, # raises false positives in beam
 
 [FORMAT]
 # Allow some single and two letter variables, but keep overall invalid-name rule
-good-names=i,j,k,f,ex,e,_,p,ip,cm
+good-names=i,j,k,f,ex,e,_,p,ip,cm,censoredplanet-analysis
 # Don't require docstring for short code segments
 docstring-min-length=10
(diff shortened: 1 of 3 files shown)


setup.cfg
```diff
@@ -30,6 +30,6 @@ disable=abstract-method, # raises false positives in beam
 
 [FORMAT]
 # Allow some single and two letter variables, but keep overall invalid-name rule
-good-names=i,j,k,f,ex,e,_,p,ip,cm
+good-names=i,j,k,f,ex,e,_,p,ip,cm,censoredplanet-analysis
```
ohnorobo, line 33: add an explanatory comment here similar to https://github.com/censoredplanet/censoredplanet-analysis/pull/113/commits/dfeefcfb2c48527aec9132a9b9c7309fa86228a9

ohnorobo (changes requested):
You should add a test that exercises the failure mode here.

ohnorobo (approved):

PR #115, opened 2022-01-19
update query to accommodate multiple errors

ohnorobo:
Parse multiple mismatch error strings in the error data.

Fortunately the errors are already ordered by importance (with certs errors first, body errors last) so we can just take the first one.

This introduces the new outcome `content/header_mismatch`.

In the future we may want to change the format in the base table to accommodate the multiple errors in a structured way, but for now this should fix the dashboard.

TESTED: ran on the prod data and saw no unknowns.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -30,6 +30,7 @@ disable=abstract-method, # raises false positives in beam
 
 [FORMAT]
 # Allow some single and two letter variables, but keep overall invalid-name rule
-good-names=i,j,k,f,ex,e,_,p,ip,cm
+# Also allow censoredplanet-analysis as a module name
+good-names=i,j,k,f,ex,e,_,p,ip,cm,censoredplanet-analysis
 # Don't require docstring for short code segments
 docstring-min-length=10
--- a/table/queries/merged_reduced_scans.sql
+++ b/table/queries/merged_reduced_scans.sql
@@ -118,6 +118,11 @@ CREATE TEMP FUNCTION ClassifyError(error STRING,
     WHEN (error = "Certificates do not match") THEN "content/tls_mismatch" # HTTPS
     WHEN (error = "Cipher suites do not match") THEN "content/tls_mismatch" # HTTPS
     WHEN (error = "TLS versions do not match") THEN "content/tls_mismatch" # HTTPS
+    # Hyperquack v2 multiple errors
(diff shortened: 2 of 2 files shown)


setup.cfg
```diff
@@ -30,6 +30,7 @@ disable=abstract-method, # raises false positives in beam
 
 [FORMAT]
 # Allow some single and two letter variables, but keep overall invalid-name rule
-good-names=i,j,k,f,ex,e,_,p,ip,cm
+# Also allow censoredplanet-analysis as a module name
+good-names=i,j,k,f,ex,e,_,p,ip,cm,censoredplanet-analysis
```
fortuna, line 34: Not used, revert?

Also, prefer underscores: `censoredplanet_analysis`, otherwise you don't have a valid symbol.

table/queries/merged_reduced_scans.sql
```diff
@@ -131,6 +136,14 @@ CREATE TEMP FUNCTION ExtractServerHeader(received_headers ARRAY<STRING>) AS (
    WHERE STARTS_WITH(header, "Server: ") LIMIT 1)
 );
 
+# Error string with either a single error or multiple split by ;
+# Returns the first error
+CREATE TEMP FUNCTION GetFirstError(errors STRING) AS (
+  IF(REGEXP_CONTAINS(errors, ';'),
+     REGEXP_EXTRACT(errors, '(.*?);'),
+     errors)
+);
```
fortuna, line 145: No need to use regex, which are more expensive.

Something like this should work:
```suggestion
  IF(STRPOS(errors, ';') > 0, SUBSTR(errors, 1, STRPOS(errors, ';') - 1), errors)
```

Reference: https://cloud.google.com/bigquery/docs/reference/standard-sql/string_functions

fortuna (changes requested):

setup.cfg
ohnorobo: This is to fix a error where it started lint triggering on the github repo name `censoredplanet-analysis`. It's too late to change that, so we'll just have to add an exception for the name.

table/queries/merged_reduced_scans.sql
ohnorobo: changed

fortuna (approved):

PR #116, opened 2022-01-31
flatten measurements for satellite

ohnorobo:
Apologies for the size of this PR 😬  I promise >1000 of these lines are just tests. Probably the best way to understand what's happening here is to start by looking at the test cases to get a sense for what outputs we're aiming for.

This change breaks out individual DNS requests into their own rows. This works in different ways for v1, v2.1, and v2.2 data since they store different information about their measurements.

The other major internal change is how we handle received ip tagging. In the old code we split up rows per-received-ip to tag, then reconstructed (grouped) them by `measurement_id`. However that makes it impossible to separate out multiple rows per measurement, so now we're using an additional `roundtrip_id` per final output row. I've also changed some intermediate stages in the code where we were storing Dicts instead of Lists in the `received` field. Now even in intermediate stages we only store lists, which matches the final output schema.

Some changes I have not made:
- The final schema format for rcodes currently is List[str], ultimately we want it to just be Int (a single rcode) but I don't want to add on that schema change here.
- I'd also like to eventually update some of the field names in the schema to be clearer
- The received ip tagging currently works in tests, but fails when running full-size backfills. I haven't been able to get this fixed yet. Currently I plan to turn off the received ip tagging temporarily in a follow-up PR so that we're able to make faster progress on the dashboard.

### DIFF
--- a/pipeline/metadata/beam_metadata.py
+++ b/pipeline/metadata/beam_metadata.py
@@ -53,10 +53,12 @@ def merge_metadata_with_rows(  # pylint: disable=unused-argument
     new_row.update(row)
     if field == 'received':
       if new_row['received']:
-        new_row['received'].update(ip_metadata)
-        new_row['received'].pop('date', None)
-        new_row['received'].pop('name', None)
-        new_row['received'].pop('country', None)
+        # Double-flattened rows are stored with a single received ip in each list
+        # to be reconstructed later
+        new_row['received'][0].update(ip_metadata)
+        new_row['received'][0].pop('date', None)
+        new_row['received'][0].pop('name', None)
+        new_row['received'][0].pop('country', None)
     else:
       new_row.update(ip_metadata)
     yield new_row
--- a/pipeline/metadata/flatten_satellite.py
+++ b/pipeline/metadata/flatten_satellite.py
@@ -76,12 +76,13 @@ def format_timestamp(timestamp: str) -> str:
   return f'{date}T{time}{timezone_hour}:{timezone_seconds}'
 
 
-def _process_received_ips_v1(
-    row: Row, received_ips: Optional[Dict[str, List[str]]]) -> Iterator[Row]:
+def _annotate_received_ips_v1(
+    base_response: Row,
+    received_ips: Optional[Dict[str, List[str]]]) -> Iterator[Row]:
   """Add received_ip metadata to a Satellite row
 
   Args:
-    row: existing row of satellite data
+    base_response: existing row of satellite data
     received_ips: data on the response from the resolver
       ex: {<ip>: ["tags"]}
 
@@ -89,8 +90,8 @@ def _process_received_ips_v1(
     Row with an additional 'received' array field
   """
   if received_ips is None:
-    row['received'] = []
-    yield row
+    base_response['received'] = []
+    yield base_response
     return
 
   # Convert to dict {<ip>: ["tags"]} with no tags
@@ -107,8 +108,8 @@ def _process_received_ips_v1(
       received['matches_control'] = _append_tags(tags)
     all_received.append(received)
 
-  row['received'] = all_received
-  yield row
+  base_response['received'] = all_received
+  yield base_response
 
 
 def _append_tags(tags: List[str]) -> str:
@@ -149,29 +150,29 @@ def split_rcodes(rcodes: List[str]) -> Tuple[List[str], List[str]]:
   return successful_test_rcode, failed_test_rcodes
 
 
-def _process_satellite_v2p1(row: Row, scan: Any,
+def _process_satellite_v2p1(base_response: Row, scan: Any,
                             filepath: str) -> Iterator[Row]:
   """Finish processing a line of Satellite v2.1 data.
 
   Args:
-    row: a partially processed row of satellite data
+    base_response: a partially processed row of satellite data
     scan: a loaded json object containing the parsed content of the line
     filepath: path like "<path>/<filename>.json.gz"
 
   Yields:
     Rows
   """
-  row['controls_failed'] = not scan['passed_control']
-  received_ips = scan.get('response').copy()
+  base_response['controls_failed'] = not scan['passed_control']
+  input_response_data = scan.get('response').copy()
 
-  if not received_ips:
-    row['received'] = []
-    yield row
+  if not input_response_data:
+    base_response['received'] = []
+    yield base_response
     return
 
   # separate into one answer ip per row for tagging
-  rcodes = received_ips.pop('rcode', [])
-  errors = received_ips.pop('error', [])
+  rcodes = input_response_data.pop('rcode', [])
+  errors = input_response_data.pop('error', [])
 
   successful_test_rcode, failed_test_rcodes = split_rcodes(rcodes)
 
@@ -180,7 +181,7 @@ def _process_satellite_v2p1(row: Row, scan: Any,
         f"Satellite v2.1 measurement has multiple 0 rcodes: {filepath} - {scan}"
     )
 
-  if not successful_test_rcode and received_ips:
+  if not successful_test_rcode and input_response_data:
     # TODO figure out how to handle this bug case from 2021-05-16
     # for now drop the measurements
     logging.warning(
@@ -190,32 +191,32 @@ def _process_satellite_v2p1(row: Row, scan: Any,
 
   if successful_test_rcode:
     all_received = []
-    for (ip, tags) in received_ips.items():
+    for (ip, tags) in input_response_data.items():
       received = {'ip': ip}
       if tags:
         received['matches_control'] = _append_tags(tags)
       all_received.append(received)
 
-    success_row = row.copy()
-    success_row['received'] = all_received
-    success_row['rcode'] = successful_test_rcode
-    yield success_row
+    success_observation = base_response.copy()
+    success_observation['received'] = all_received
+    success_observation['rcode'] = successful_test_rcode
+    yield success_observation
 
   for rcode in failed_test_rcodes:
-    failed_row = row.copy()
-    failed_row['rcode'] = [rcode]
+    failed_observation = base_response.copy()
+    failed_observation['rcode'] = [rcode]
     # -1 rcodes corrospond to the error messages in order.
     if rcode == "-1":
-      failed_row['error'] = errors.pop() if errors else None
-    yield failed_row
+      failed_observation['error'] = errors.pop() if errors else None
+    yield failed_observation
 
   # There is a bug in some v2.1 data where -1 rcodes weren't recorded.
   # In that case we add them back in manually
   for error in errors:
-    error_row = row.copy()
-    error_row['rcode'] = ["-1"]
-    error_row['error'] = error
-    yield error_row
+    error_observation = base_response.copy()
+    error_observation['rcode'] = ["-1"]
+    error_observation['error'] = error
+    yield error_observation
 
 
 class SatelliteFlattener():
@@ -297,7 +298,7 @@ class SatelliteFlattener():
       row['error'] = json.dumps(row['error'])
 
     received_ips = scan.get('answers')
-    yield from _process_received_ips_v1(row, received_ips)
+    yield from _annotate_received_ips_v1(row, received_ips)
 
   def _process_satellite_v2(self, scan: Any, filepath: str,
                             random_measurement_id: str) -> Iterator[Row]:
--- a/pipeline/metadata/flatten_satellite.py
+++ b/pipeline/metadata/flatten_satellite.py
@@ -14,6 +14,9 @@ from pipeline.metadata.flatten_base import Row
 from pipeline.metadata.blockpage import BlockpageMatcher
 from pipeline.metadata.domain_categories import DomainCategoryMatcher
 
+# Type definition for input responses, TODO make actual type stricter
+ResponsesEntry = Any
+
 SATELLITE_TAGS = {'ip', 'http', 'asnum', 'asname', 'cert'}
 SATELLITE_V2_1_START_DATE = datetime.date(2021, 3, 1)
 SATELLITE_V2_2_START_DATE = datetime.date(2021, 6, 24)
@@ -150,20 +153,20 @@ def split_rcodes(rcodes: List[str]) -> Tuple[List[str], List[str]]:
   return successful_test_rcode, failed_test_rcodes
 
 
-def _process_satellite_v2p1(base_response: Row, scan: Any,
+def _process_satellite_v2p1(base_response: Row, responses_entry: ResponsesEntry,
                             filepath: str) -> Iterator[Row]:
   """Finish processing a line of Satellite v2.1 data.
 
   Args:
     base_response: a partially processed row of satellite data
-    scan: a loaded json object containing the parsed content of the line
+    responses_entry: a loaded json object containing the parsed content of the line
     filepath: path like "<path>/<filename>.json.gz"
 
   Yields:
     Rows
   """
-  base_response['controls_failed'] = not scan['passed_control']
-  input_response_data = scan.get('response').copy()
+  base_response['controls_failed'] = not responses_entry['passed_control']
+  input_response_data = responses_entry.get('response').copy()
 
   if not input_response_data:
     base_response['received'] = []
@@ -178,14 +181,14 @@ def _process_satellite_v2p1(base_response: Row, scan: Any,
 
   if any(rcode == "0" for rcode in failed_test_rcodes):
     raise Exception(
-        f"Satellite v2.1 measurement has multiple 0 rcodes: {filepath} - {scan}"
+        f"Satellite v2.1 measurement has multiple 0 rcodes: {filepath} - {responses_entry}"
     )
 
   if not successful_test_rcode and input_response_data:
     # TODO figure out how to handle this bug case from 2021-05-16
     # for now drop the measurements
     logging.warning(
-        f"Satellite v2.1 measurement has ips but no 0 rcode: {filepath} - {scan}"
+        f"Satellite v2.1 measurement has ips but no 0 rcode: {filepath} - {responses_entry}"
     )
     return
 
@@ -227,13 +230,13 @@ class SatelliteFlattener():
     self.blockpage_matcher = blockpage_matcher
     self.category_matcher = category_matcher
 
-  def process_satellite(self, filepath: str, scan: Any,
+  def process_satellite(self, filepath: str, responses_entry: ResponsesEntry,
                         random_measurement_id: str) -> Iterator[Row]:
     """Process a line of Satellite data.
 
     Args:
       filepath: a filepath string
-      scan: a loaded json object containing the parsed content of the line
+      responses_entry: a loaded json object containing the parsed content of the line
       random_measurement_id: a hex id identifying this individual measurement
 
     Yields:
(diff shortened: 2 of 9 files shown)


pipeline/metadata/flatten_satellite.py
```diff
@@ -75,111 +76,146 @@ def format_timestamp(timestamp: str) -> str:
...
 
 
-def _process_received_ips(
-    row: Row, received_ips: Optional[Dict[str, str]]) -> Iterator[Row]:
+def _process_received_ips_v1(
+    row: Row, received_ips: Optional[Dict[str, List[str]]]) -> Iterator[Row]:
   """Add received_ip metadata to a Satellite row
+
   Args:
     row: existing row of satellite data
```
fortuna, line 84: Is `row` what we read from their file? What does it represent?

pipeline/metadata/flatten_satellite.py
```diff
@@ -75,111 +76,146 @@ def format_timestamp(timestamp: str) -> str:
...
-              [tag for tag in matched if tag in SATELLITE_TAGS])
-        row['received'].append(received)
-      yielded = True
-      yield row.copy()
-    # Do not output rows if there are no trials for the test domain.
-  if not yielded and has_test_domain:
-    # All trials for the test domain in `scan` are errors
-    yield row.copy()
+  row['controls_failed'] = not scan['passed_control']
+  received_ips = scan.get('response').copy()
```
fortuna, line 165: Rename `received_ips` to `response`.

pipeline/metadata/flatten_satellite.py
```diff
@@ -75,111 +76,146 @@ def format_timestamp(timestamp: str) -> str:
...
+
+  if not received_ips:
+    row['received'] = []
+    yield row
+    return
+
+  # separate into one answer ip per row for tagging
+  rcodes = received_ips.pop('rcode', [])
```
fortuna, line 173: Why would you have more than one rcode and error in a response?

This is very hard to follow. I propose a logic change to first:
- extract all responses from the Satellite data. Each response must only have one rcode and error.
- Annotate and output the responses.

The first step changes with the data version, but the second should be shared across versions.

Introduce functions like get_responses_v1 for example. That may even be its own step in the pipeline.

pipeline/metadata/flatten_satellite.py
```diff
@@ -371,54 +463,60 @@ def _process_satellite_v2_control(
...
-              format_timestamp(responses[-1]['end_time']),
+              format_timestamp(response['end_time']),
           'anomaly':
               None,
           'success':
               not scan['connect_error'],
           'controls_failed':
               not scan['passed_control'],
-          'rcode': [str(response['rcode']) for response in responses],
+          'rcode': [str(response['rcode'])],
```
fortuna, line 494: This should not be a list.

pipeline/metadata/flatten_satellite.py
```diff
@@ -371,54 +463,60 @@ def _process_satellite_v2_control(
...
-      # always contains some non-control trial domains
-      is_control_domain = False
+
+    for response in responses:
+      is_control_domain = flatten_base.is_control_url(response['url'])
 
       row = {
           'domain':
-              scan['test_url'],
+              response['url'],
```
fortuna, line 472: Extract the domain from the URL.
Why is the field called URL?

pipeline/metadata/flatten_satellite.py
fortuna: The flow should be:
input entry -> responses -> annotated responses
Having the data for multiple responses in one merged output row is not analyzable in a sane way. We can't do that.

fortuna (changes requested):

pipeline/metadata/flatten_satellite.py
```diff
@@ -75,111 +76,146 @@ def format_timestamp(timestamp: str) -> str:
...
+    received_ips: data on the response from the resolver
+      ex: {<ip>: ["tags"]}
+
   Yields:
-    Rows
+    Row with an additional 'received' array field
   """
-  if not received_ips:
+  if received_ips is None:
+    row['received'] = []
```
fortuna, line 92: Rename `row` to `base_response`.

pipeline/metadata/flatten_satellite.py
fortuna: Actually, maybe we should rename `row` to `base_observation` and `received_ips` to something like `input_response_data`, since it's not a response, and it's not just IPs.

I'm using "observation", instead of "responses" because sometimes the observation is not a response (e.g. timeout)

pipeline/metadata/flatten_satellite.py
```diff
@@ -75,111 +76,146 @@ def format_timestamp(timestamp: str) -> str:
...
+
+  if successful_test_rcode:
+    all_received = []
+    for (ip, tags) in received_ips.items():
+      received = {'ip': ip}
+      if tags:
+        received['matches_control'] = _append_tags(tags)
+      all_received.append(received)
+
+    success_row = row.copy()
```
fortuna, line 199: Rename `success_row` to `observation`. It should represent a single DNS request + observation.

pipeline/metadata/flatten_satellite.py
```diff
@@ -75,111 +76,146 @@ def format_timestamp(timestamp: str) -> str:
...
+
+  if any(rcode == "0" for rcode in failed_test_rcodes):
+    raise Exception(
+        f"Satellite v2.1 measurement has multiple 0 rcodes: {filepath} - {scan}"
+    )
+
+  if not successful_test_rcode and received_ips:
+    # TODO figure out how to handle this bug case from 2021-05-16
+    # for now drop the measurements
+    logging.warning(
```
fortuna, line 186: Perhaps introduce an error counter

pipeline/metadata/flatten_satellite.py
```diff
@@ -75,111 +76,146 @@ def format_timestamp(timestamp: str) -> str:
...
+        f"Satellite v2.1 measurement has ips but no 0 rcode: {filepath} - {scan}"
+    )
+    return
+
+  if successful_test_rcode:
+    all_received = []
+    for (ip, tags) in received_ips.items():
```
fortuna, line 193: Isn't received_ips a dict with fields like "rcode"?

pipeline/metadata/flatten_satellite.py
```diff
@@ -75,111 +76,146 @@ def format_timestamp(timestamp: str) -> str:
...
+        received['matches_control'] = _append_tags(tags)
+      all_received.append(received)
+
+    success_row = row.copy()
+    success_row['received'] = all_received
+    success_row['rcode'] = successful_test_rcode
+    yield success_row
+
+  for rcode in failed_test_rcodes:
+    failed_row = row.copy()
```
fortuna, line 205: I'd call this `observation`

pipeline/metadata/flatten_satellite.py
```diff
@@ -75,111 +76,146 @@ def format_timestamp(timestamp: str) -> str:
...
+    failed_row['rcode'] = [rcode]
+    # -1 rcodes corrospond to the error messages in order.
+    if rcode == "-1":
+      failed_row['error'] = errors.pop() if errors else None
+    yield failed_row
+
+  # There is a bug in some v2.1 data where -1 rcodes weren't recorded.
+  # In that case we add them back in manually
+  for error in errors:
+    error_row = row.copy()
```
fortuna, line 215: Rename to `observation`. In this case we can't really call `response`, because there was no dns response. (though perhaps you got an ICMP error)

pipeline/metadata/flatten_satellite.py
```diff
@@ -75,111 +76,146 @@ def format_timestamp(timestamp: str) -> str:
...
+  if successful_test_rcode:
+    all_received = []
+    for (ip, tags) in received_ips.items():
+      received = {'ip': ip}
+      if tags:
+        received['matches_control'] = _append_tags(tags)
+      all_received.append(received)
+
+    success_row = row.copy()
+    success_row['received'] = all_received
```
fortuna, line 200: Rename `received` to `answer`. The rcode you also "received".

pipeline/metadata/flatten_satellite.py
```diff
@@ -251,7 +287,7 @@ def _process_satellite_v1(self, date: str, scan: Any, filepath: str,
         'error': scan.get('error', None),
         'anomaly': not scan['passed'] if 'passed' in scan else None,
         'success': 'error' not in scan,
-        'received': None,
+        'received': [],
```
fortuna, line 290: `answer`

fortuna (changes requested):

pipeline/metadata/flatten_satellite.py
ohnorobo: That will be changed in https://github.com/censoredplanet/censoredplanet-analysis/pull/117, but not this PR.

pipeline/metadata/flatten_satellite.py
ohnorobo: schema name changes will happen in https://github.com/censoredplanet/censoredplanet-analysis/pull/119

pipeline/metadata/flatten_satellite.py
ohnorobo: https://github.com/censoredplanet/censoredplanet-analysis/pull/119

pipeline/metadata/flatten_satellite.py
```diff
@@ -75,111 +76,146 @@ def format_timestamp(timestamp: str) -> str:
...
+
   Args:
     row: existing row of satellite data
-    received_ips: data on the response from the resolver (error or ips)
+    received_ips: data on the response from the resolver
+      ex: {<ip>: ["tags"]}
+
   Yields:
-    Rows
+    Row with an additional 'received' array field
```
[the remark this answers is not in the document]
ohnorobo: I agree that would be very helpful, but that's a big refactor that can't happen in this PR.

pipeline/metadata/flatten_satellite.py
ohnorobo: Changed to `error_observation`

pipeline/metadata/flatten_satellite.py
ohnorobo: Changed to `failed_observation`

pipeline/metadata/flatten_satellite.py
ohnorobo: In v2.1 the response field looks like this:
3hksuhqzaqs2h7k and contains results from several roundtrip tests.

https://censoredplanet.slack.com/archives/C01DK5FF4KD/p1639594115136600

So we need to untangle all those fields into rows where each represents an individual roundtrip. First we pop off the rcode/error fields, then we treat the remainder as a dict {'ip': ['tag_a', 'tag_b']}

pipeline/metadata/flatten_satellite.py
ohnorobo: We can count these logs in a job that succeeded. There are 4070 rows that trigger this error in the whole dataset.

6fGL4r5iN8nsVjf

https://console.cloud.google.com/logs/query;query=resource.type%3D%22dataflow_step%22%0Aresource.labels.job_id%3D%222022-01-17_20_58_59-987126952126004496%22%0Aseverity%3DWARNING%0A%22Satellite%20v2.1%20measurement%20has%20ips%20but%20no%200%20rcode%22;timeRange=2022-01-18T04:58:59.933Z%2F2022-01-18T13:01:53.036Z;cursorTimestamp=2022-01-18T11:06:33.445386171Z?project=firehook-censoredplanet

We don't expect this number to ever get any larger (since no more 2.1 data is being generated.) So this number will never get bigger. Implementing a counter in code doesn't seem like a good use of time.

pipeline/metadata/flatten_satellite.py
ohnorobo: renamed to `success_observation`. That is what it represents.

ohnorobo: renamed `row` to `base_observation` and `received_ips` to `input_response_data`

pipeline/metadata/flatten_satellite.py
ohnorobo: done

pipeline/metadata/flatten_satellite.py
ohnorobo: renamed `row` to `base_observation` and `received_ips` to `input_response_data`

pipeline/metadata/flatten_satellite.py
```diff
@@ -75,111 +76,146 @@ def format_timestamp(timestamp: str) -> str:
   return f'{date}T{time}{timezone_hour}:{timezone_seconds}'
 
 
-def _process_received_ips(
-    row: Row, received_ips: Optional[Dict[str, str]]) -> Iterator[Row]:
+def _process_received_ips_v1(
```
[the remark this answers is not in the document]
ohnorobo: renamed to `_annotate_received_ips_v1`

pipeline/metadata/beam_metadata.py
```diff
@@ -53,10 +53,12 @@ def merge_metadata_with_rows(  # pylint: disable=unused-argument
     new_row.update(row)
     if field == 'received':
       if new_row['received']:
-        new_row['received'].update(ip_metadata)
-        new_row['received'].pop('date', None)
-        new_row['received'].pop('name', None)
-        new_row['received'].pop('country', None)
+        # Double-flattened rows are stored with a single received ip in each list
+        # to be reconstructed later
+        new_row['received'][0].update(ip_metadata)
```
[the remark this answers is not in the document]
ohnorobo: The way we're doing the received ip tagging is:
-  there's a flattened (representing a single observation) row with multiple received ips. 
- We then "double flatten" that into several rows each with a single received ip each (and the same `roundtrip_id`)
- we then add tags and do the co-group with metadata. (the co-group is why this whole dance is required)
- Then we reconstruct the original measurement with multiple received_ips [like this](https://github.com/censoredplanet/censoredplanet-analysis/blob/fb0d8a5b4dfa90edeca2347c7141f0d5f331c085/pipeline/metadata/satellite.py#L376).

We may want to make some changes here, but I didn't want to do that while also doing the flattening. If the row has made it to this point in the code its `received` field will be a list with a single ip like [{'ip': '1.2.3.4'}] ready to have metadata added.

pipeline/metadata/flatten_satellite.py
ohnorobo: `row` is the python object we are constructing in order to write it as a row in bigquery. Data read from the scan file is usually called `line` (if it's a json string) or `scan` if we've parsed the json into a python object.

pipeline/metadata/flatten_satellite.py
ohnorobo: I'm not merging multiple responses into one row here. I'm trying to tease apart multiple responses that are already merged in the scan data into individual rows which each represent a single observation.

I do agree that this code is hard to understand though. Maybe we should get on a call and do a code walkthrough to see where there's bad flow/lack of needed comments explaining the data format?

pipeline/metadata/flatten_satellite.py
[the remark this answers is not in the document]
ohnorobo: The options seem to be using either NamedTuple or the new [dataclass](https://www.python.org/dev/peps/pep-0557/)

pipeline/metadata/flatten_satellite.py
```diff
@@ -371,54 +464,60 @@ def _process_satellite_v2_control(
         Rows
     """
     responses = scan.get('response', [])
```
fortuna, line 466: Please rename `scan` to `responses_entry`. Introduce a `ResponsesEntry` type that corresponds to the responses.json and responses_control.json files (as defined at https://docs.censoredplanet.org/dns.html)

pipeline/metadata/flatten_satellite.py
```diff
@@ -371,54 +464,60 @@ def _process_satellite_v2_control(
...
           'start_time':
-              format_timestamp(responses[0]['start_time']),
+              format_timestamp(response['start_time']),
           'end_time':
-              format_timestamp(responses[-1]['end_time']),
+              format_timestamp(response['end_time']),
           'anomaly':
               None,
           'success':
               not scan['connect_error'],
```
fortuna, line 492: I don't see `connect_error` and 'passed_control` in the `responses.json` type definition.

pipeline/metadata/flatten_satellite.py
fortuna: Also specify in the comment the filename this is coming from.

pipeline/metadata/flatten_satellite.py
```diff
@@ -371,54 +464,60 @@ def _process_satellite_v2_control(
...
-              row['received'] = {'ip': ip}
-              yield row.copy()
+
+      if response['error'] == 'null':
+        error = None
+      else:
+        error = response['error']
+      row['error'] = error
+
+      received_ips = response['response']
```
fortuna, line 510: I don't see `response` in `ResponsesEntry.response`.  Did you mean `answer`?

fortuna (changes requested):

pipeline/metadata/flatten_satellite.py
ohnorobo: This code is heavily unit tested (with the unit tests coming from example measurements) and has been run against the entire satellite dataset. If you're seeing fields that don't match the documentation it's probably that the docs are wrong.

Here's the example measurement in the unit tests with this field https://github.com/censoredplanet/censoredplanet-analysis/blob/16276c030dc9482c3c0601850f76503001c36cb4/pipeline/metadata/test_satellite.py#L560

pipeline/metadata/flatten_satellite.py
ohnorobo: I'm happy to do either of those name changes. But both of them require a schema change, so they have to happen in https://github.com/censoredplanet/censoredplanet-analysis/pull/119.

Currently, I need to preserve the full content of this field so I'll be able to investigate it and figure out why these are appearing as urls at all. If I do the extraction here (in this PR) then that data will become impossible to access.

fortuna: I don't see how adding a column would be a problem. It actually concerns me that we may not be aggregating correctly if it's urls and not domains.
I'm fine with punting on that for now, but please add a TODO

pipeline/metadata/flatten_satellite.py
ohnorobo: Same as above. Here are the example test with the `response.response` field

https://github.com/censoredplanet/censoredplanet-analysis/blob/534652d4ea0209d49a176e1c9ddbee5bfeae7a67/pipeline/metadata/test_flatten_satellite.py#L189

pipeline/metadata/flatten_satellite.py
ohnorobo: done and done

pipeline/metadata/satellite.py
```diff
@@ -396,14 +396,52 @@ def _unflatten_satellite(flattened_measurement: Iterable[Row]) -> Iterator[Row]:
...
     # during the tagging step if a vantage point also appears as a response IP.
     for tag in ['asname', 'asnum', 'http', 'cert']:
       combined.pop(tag, None)
+
+    combined.pop('roundtrip_id')  # Remove roundtrip id
     yield combined
 
 
+def flatten_received_ips(row: Row) -> Iterator[Row]:
+  """Flatten a row with multiple received ips into rows with a single ip.
```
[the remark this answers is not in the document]
ohnorobo: Changed the name to dns_roundtrip

changed the <dict> abstraction in the comments to be {'ip': "1.2.3.4'} which matches the structure of the actual data.

pipeline/metadata/flatten_satellite.py
ohnorobo: TODO added

pipeline/metadata/satellite.py
```diff
@@ -396,14 +396,60 @@ def _unflatten_satellite(flattened_measurement: Iterable[Row]) -> Iterator[Row]:
...
+  dns_roundtrip['roundtrip_id'] = uuid.uuid4().hex
+
+  all_received = dns_roundtrip['received']
+
+  if len(all_received) == 0:
+    yield dns_roundtrip
+    return
+
+  for received in all_received:
+    dns_roundtrip['received'] = [received]
```
fortuna, line 449: This is type abuse. The new field is no longer a list, and the structure has a different semantics. You need to define a new type. But see my refactoring suggestion above. By not carrying the entire observation around, you are forced to not reuse the type.

pipeline/metadata/satellite.py
```diff
@@ -396,14 +396,60 @@ def _unflatten_satellite(flattened_measurement: Iterable[Row]) -> Iterator[Row]:
...
+            'ip': '1.2.3.4'
+        },{
+            'ip': '5.6.7.8'
+        }]
+      }
+
+  Returns:
+    Rows like
+      {
+        "field": "value"
```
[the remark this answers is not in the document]
fortuna: BTW, the transform doesn't have to output two collections. You can run two separate transforms on the same input instead, which is cleaner.

So one transform would do:
dsn_roundtrips -> (roundtrip_id, dns_roundtrip)

and the other would do:
dns_roundtrips -> (roundtrip_id, answer_ip)

pipeline/metadata/satellite.py
```diff
@@ -421,13 +467,19 @@ def process_satellite_with_tags(
   rows = (
       row_lines | 'flatten json' >> beam.ParDo(
           flatten.FlattenMeasurement()).with_output_types(Row))
+
+  # PCollection[Row] each with only a single element in the received arrays
+  received_ip_flattened_rows = (
+      rows | 'flatten received ips' >>
```
fortuna, line 473: This is where you can create the two separate PCollections by running two different transforms on the `rows`.

fortuna (changes requested):

fortuna (approved):
Let's add a TODO for the triple CoGroup

ohnorobo:
Added a TODO about the triple cogroup in [8de0daa](https://github.com/censoredplanet/censoredplanet-analysis/pull/116/commits/8de0daaffeee1f4da56a1c33a2698222327bd03b)

As discussed offline I'm going to merge this now and work on some incremental refactoring in future PRs.

PR #119, opened 2022-02-03
update schema names

ohnorobo:
Change satellite schema names to match [this document](https://docs.google.com/document/d/1BMohMUiqJzYnTR8fdEqb1Ang5r1oW_aEHdsxQraQPR8/edit).

I'm updating the names in:
- the bigquery schemas
- the `flatten_for_bigquery_` translation layer
- in the sql query
- in the documentation

I'm not updating the names in the dataclasses themselves, since that would involve changing everywhere they're used in the code. I'll do that in a follow-up where it makes sense.


pipeline/beam_tables.py
```diff
@@ -123,7 +123,7 @@ def _add_schemas(schema_a: Dict[str, Any],
     BASE_BIGQUERY_SCHEMA, {
         'name': ('string', 'nullable'),
```
fortuna, line 124: What's `name`?

pipeline/beam_tables.py
```diff
@@ -123,7 +123,7 @@ def _add_schemas(schema_a: Dict[str, Any],
     BASE_BIGQUERY_SCHEMA, {
```
fortuna, line 123: We may want to discuss the schema in a doc instead, since we can't comment on the base schema and explore suggestions here easily.

pipeline/beam_tables.py
```diff
@@ -123,7 +123,7 @@ def _add_schemas(schema_a: Dict[str, Any],
...
         'name': ('string', 'nullable'),
         'is_control_ip': ('boolean', 'nullable'),
-        'received': ('record', 'repeated', {
```
fortuna, line 126: Group `answer` and `rcode` in a response record to convey the right semantics.

pipeline/beam_tables.py
ohnorobo: Yes. I didn't make this intending to discuss the schema here, it was just a placeholder for the future.

docs/base_tables.md
```diff
@@ -28,50 +28,51 @@ The json data is processed into a flat table format which looks like this.
...
 |                           |
 | success                   | BOOLEAN      | Did the individual roundtrip measurement succeed? |
 | anomaly                   | BOOLEAN      | Was interference detected in the overall measurement? |
 | stateful_block            | BOOLEAN      | Was stateful interference detected? |
-| controls_failed            | BOOLEAN     | Did all the control measurements connected to this measurement fail? |
+| controls_failed           | BOOLEAN     | Did all the control measurements connected to this measurement fail? |
 |                           |
 | **Internal**              |
 |                           |
+| measurement_id            | STRING       | A uuid which is the same for observations which are part of the same measurement. </br> If there are 5 retries of a scan they will all have the same id. </br> eg. `a08df2fe70d54092916b8df87e330f47` |
```
fortuna, line 75:
```suggestion
| measurement_id            | STRING       | A unique id which is the same for observations which are part of the same measurement. </br> If there are 5 retries of a scan they will all have the same id. </br> eg. `a08df2fe70d54092916b8df87e330f47` |
```
It doesn't have to be a uuid

pipeline/metadata/schema.py
```diff
@@ -250,35 +250,35 @@ def flatten_for_bigquery_satellite(row: SatelliteRow) -> Dict[str, Any]:
...
+      'domain_category': row.category,
+      'is_control_domain': row.is_control,
       'date': row.date,
       'start_time': row.start_time,
       'end_time': row.end_time,
-      'error': row.error,
-      'anomaly': row.anomaly,
+      'resolver_ip': row.ip,
+      'resolver_name': row.ip_metadata.name,
+      'is_control_resolver': row.is_control_ip,
```
fortuna, line 260: I'd use `resolver_is_trusted`.
`control` isn't clear, and better to have the `resolver_` prefix on all resolver fields.

pipeline/metadata/schema.py
```diff
@@ -250,35 +250,35 @@ def flatten_for_bigquery_satellite(row: SatelliteRow) -> Dict[str, Any]:
...
+      'received_rcode': row.rcode,
+      'answers': [],
       'success': row.success,
-      'is_control': row.is_control,
-      'controls_failed': row.controls_failed,
-      'measurement_id': row.measurement_id,
-      'source': row.source,
-      'is_control_ip': row.is_control_ip,
-      'rcode': row.rcode,
+      'anomaly': row.anomaly,
```
fortuna, line 272: `legacy_anomaly`?

docs/base_tables.md
```diff
@@ -28,50 +28,51 @@ The json data is processed into a flat table format which looks like this.
...
 |                           |
 | date                      | DATE         | Date that an individual measurement was taken |
 | start_time                | TIMESTAMP    | Start time of the individual measurement |
 | end_time                  | TIMESTAMP    | End time of the individual measurement |
-| measurement_id            | STRING       | A uuid which is the same for observations which are part of the same measurement. </br> If there are 5 retries of a scan they will all have the same id. </br> eg. `a08df2fe70d54092916b8df87e330f47` |
-| error                     | STRING       | Any error, eg. `Network Timeout` |
+|                           |
+| **Vantage Point Server**  |
+|                           |
+| server_ip                 | STRING       | The ip address of the server being tested, eg. `1.1.1.1` |
```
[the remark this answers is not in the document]
ohnorobo: no, there's no documentation for the satellite base table yet since we're not publishing it in the base dataset.

pipeline/metadata/schema.py
```diff
@@ -250,35 +250,35 @@ def flatten_for_bigquery_satellite(row: SatelliteRow) -> Dict[str, Any]:
...
+      'domain_controls_failed': row.controls_failed,
       'average_confidence': row.average_confidence,
       'untagged_controls': row.untagged_controls,
       'untagged_response': row.untagged_response,
       'excluded': row.excluded,
       'exclude_reason': row.exclude_reason,
       'has_type_a': row.has_type_a,
```
eltsai, line 279: I think we can remove the `has_type_a` field (indicating if there's at least one `rcode=0` and `ip != null` response for a DNS query), since we already have a `rcode` field

docs/base_tables.md
```diff
@@ -28,50 +28,51 @@ The json data is processed into a flat table format which looks like this.
 | **Measured Domain**       |
 |                           |
 | domain                    | STRING       | The domain being tested, eg. `example.com` |
-| is_control                | STRING       | If the measured domain a control domain? |
-| category                  | STRING       | The [category](domain_categories.md) of the domain being tested, eg. `Social Networking`, `None` if unknown |
+| is_control_domain         | STRING       | If the measured domain a control domain? |
```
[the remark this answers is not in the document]
ohnorobo: I thought about this too, but I think 
`IF table_name.is_control_domain` 
reads a little more cleanly than 
`IF table_name.domain_is_control`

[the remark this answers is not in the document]
fortuna: I find the first is actually more readable, as it clearly expresses the relationship to the domain concept, even though it's not proper English. is_control_domain is inconsistent with all the other names.

But I'll leave it up to you.

docs/base_tables.md
fortuna: Not important. It's just a promise you don't need to make.

pipeline/metadata/schema.py
fortuna: What about this?

fortuna (approved):

docs/base_tables.md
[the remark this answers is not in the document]
ohnorobo: switched to domain_is_control

pipeline/metadata/schema.py
ohnorobo: switched to resolver_is_trusted

PR #122, opened 2022-02-16
set controls_failed to False in satv1 data

ohnorobo:
This will make the fields more consistent over different versions and let us remove an additional check [here](https://github.com/censoredplanet/censoredplanet-analysis/blob/7a1b3bea1472e38336761e39fa28ac012ca4f01e/table/queries/derived_satellite_scans.sql#L134).


pipeline/metadata/flatten_satellite.py
```diff
@@ -287,6 +287,8 @@ def _process_satellite_v1(self, date: str, responses_entry: ResponsesEntry,
             responses_entry['query'],
         'is_control':
             False,  # v1 doesn't have domain controls
+        'controls_failed':
+            False,  # v1 doesn't have domain controls
```
fortuna, line 291: If we don't have a control, shouldn't we assume it failed? Its absence sounds equivalent to a failure.

ohnorobo: We filter out measurements with failed controls. So if we made this True we would filter out 100% of the sat v1 measurements.

The field is not "did a control succeed", it's "did any related controls fail"

so running 0 controls -> 0 controls failed -> False

fortuna: What control are we talking about? The control resolver or the control domain? We need better names... Perhaps some renaming early would indeed help. It doesn't have to be the final names. We should be able to rename easily.

There's also a data modeling issue here. Why are we merging control and non control data?

fortuna (approved):

ohnorobo:
Renaming fields is easy-ish to do in code, but annoying to do in practice since it requires running a full backfill. That's why I want to do it all at once in https://github.com/censoredplanet/censoredplanet-analysis/pull/119 (and hash it out in a doc ahead of time)

PR #125, opened 2022-02-23
Simplify received ip tagging

ohnorobo:
Do received ip tagging using the "triple cogroup" algorithm we discussed.

Successful full backfill [here](https://pantheon.corp.google.com/dataflow/jobs/us-central1/2022-02-23_13_54_58-12382344337112806992?project=firehook-censoredplanet) (took 9 hours) Previously these backfills were failing due to OOMing workers.

I'm very hopeful this will help with performance since previously the unit tests took ~4 minutes to run, and now they take ~15 seconds.

Copying the TODO comment from the code for posterity:
```
TODO we want to change the structure here
Currently we're CoGrouping over two elements using a (date, received_ip) key
A row with flattened received ips like
 {"ip": "1.2.3.4",
  "date": "2021-01-01"
  "domain": "example.com"
  "roundtrip_id": "abc"
  "received" [{"ip": "5.6.7.8"}]
}
and metadata like
{"ip": "5.6.7.8", "date":"2021-01-01", "asn": 123}

But in the future we want to have three pieces
1. The initial row (without flattening the received ips) with a roundtrip_id
2. Information on a single received ip like:
    {"received_ip": "5.6.7.8", "date": "2021-01-01", "roundtrip_id": "abc"}
3. Metadata as above.
Then we can CoGroup 2 and 3 together by (date, received_ip)
and then Group them with 1 by roundtrip_id
and add metadata to the appropriate answers
This way we avoid having to multiply rows by their number of received ips
```

### DIFF
--- a/pipeline/metadata/satellite.py
+++ b/pipeline/metadata/satellite.py
@@ -46,8 +46,10 @@ SCAN_TYPE_SATELLITE = 'satellite'
 SCAN_TYPE_BLOCKPAGE = 'blockpage'
 
 CDN_REGEX = re.compile("AMAZON|Akamai|OPENDNS|CLOUDFLARENET|GOOGLE")
-NUM_DOMAIN_PARTITIONS = 250
 NUM_SATELLITE_INPUT_PARTITIONS = 3
+
+# PCollection key name used internally by the beam pipeline
+RECEIVED_IPS_PCOLLECTION_NAME = 'received_ips'
 
 
 def _merge_dicts(dicts: Iterable[Dict[Any, Any]]) -> Dict[Any, Any]:
@@ -355,18 +357,19 @@ def post_processing_satellite(
 
 
 def _merge_tagged_answers_with_rows(
-    key: str, value: Dict[str, Union[List[Row], List[List[Row]]]]  # pylint: disable=unused-argument
+    key: str,  # pylint: disable=unused-argument
+    value: Dict[str, Union[List[Row], List[List[Row]]]]
 ) -> Row:
   """
   Args:
     key: roundtrip_id, unused
     value:
-      {IP_METADATA_PCOLLECTION_NAME: One element list containing a row
-       ROWS_PCOLLECION_NAME: One element list of many element
+      {ROWS_PCOLLECION_NAME: One element list containing a row
+       RECEIVED_IPS_PCOLLECTION_NAME: One element list of many element
                              list containing tagged answer rows
       }
       ex:
-        { IP_METADATA_PCOLLECTION_NAME: [{
+        {ROWS_PCOLLECION_NAME: [{
             'ip': '1.2.3.4'
             'domain': 'ex.com'
             'received' : [{
@@ -376,7 +379,7 @@ def _merge_tagged_answers_with_rows(
               },
             ]
           }],
-        ROWS_PCOLLECION_NAME: [[{
+         RECEIVED_IPS_PCOLLECTION_NAME: [[{
             'ip': '4.5.6.7'
             'asname': 'AMAZON-AES',
             'asnum': 14618,
@@ -403,11 +406,12 @@ def _merge_tagged_answers_with_rows(
       ]
     }
   """
-  row: Row = value[IP_METADATA_PCOLLECTION_NAME][0]  # type: ignore
+  row: Row = value[ROWS_PCOLLECION_NAME][0]  # type: ignore
 
-  if len(value[ROWS_PCOLLECION_NAME]) == 0:  # No tags
+  if len(value[RECEIVED_IPS_PCOLLECTION_NAME]) == 0:  # No tags
     return row
-  tagged_answers: List[Row] = value[ROWS_PCOLLECION_NAME][0]  # type: ignore
+  tagged_answers: List[Row] = value[RECEIVED_IPS_PCOLLECTION_NAME][
+      0]  # type: ignore
 
   for untagged_answer in row['received']:
     untagged_ip = untagged_answer['ip']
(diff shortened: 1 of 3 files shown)


pipeline/beam_tables.py
```diff
@@ -485,8 +485,8 @@ def _add_metadata(
 
     # PCollection[Tuple[Tuple[date,ip],Dict[input_name_key,List[Row]]]]
     grouped_metadata_and_rows = (({
-        IP_METADATA_PCOLLECTION_NAME: ips_with_metadata,
-        ROWS_PCOLLECION_NAME: rows_keyed_by_ip_and_date
+        BEAM_COGROUP_A_SIDE: ips_with_metadata,
```
fortuna, line 488: Provide meaningful names instead of A and B.

pipeline/metadata/satellite.py
```diff
@@ -92,6 +92,46 @@ def get_blockpage_table_name(table_name: str, scan_type: str) -> str:
...
+        'ip': '5.6.7.8',
+        'date': '2021-01-01',
+        'roundtrip_id: 'abc'
+      }
+  """
+  roundtrip_id = row['roundtrip_id']
+  date = row['date']
+
+  for answer in row['received']:
+    received_with_id = answer.copy()
```
fortuna, line 129: This is abusing the received type.
Why not put the entire received data in a field, instead of expanding it in the parent dict?

pipeline/metadata/satellite.py
```diff
@@ -158,10 +198,22 @@ def _read_satellite_tags(filename: str, line: str) -> Iterator[Row]:
   yield tags
 
 
+def _is_vantage_point_tag(row: Row) -> bool:
+  """Determine if a given tag should be applied to vantage point or received ips.
+
+  TODO: remove this function and instead do the partitioning
```
fortuna, line 204: Yeah, they should not be merged in the first place. You're merging to then separate. And the merged shuffling is less performant

pipeline/metadata/satellite.py
```diff
@@ -179,8 +243,8 @@ def _add_vantage_point_tags(
 
   # PCollection[Tuple[Tuple[date,ip],Dict[input_name_key,List[Row]]]]
   grouped_metadata_and_rows = (({
-      IP_METADATA_PCOLLECTION_NAME: ips_with_metadata,
-      ROWS_PCOLLECION_NAME: rows_keyed_by_ip_and_date
+      BEAM_COGROUP_A_SIDE: ips_with_metadata,
```
fortuna, line 246: Use meaningful names that represent the data.

pipeline/metadata/satellite.py
```diff
@@ -375,86 +354,188 @@ def _flat_rows_controls(key: Any, value: Row) -> Iterator[Tuple[Row, int]]:  # p
...
+      received_ips_with_roundtrip_ip_and_date |
+      'received ip: key by ips and dates' >>
+      beam.Map(lambda row: (make_date_ip_key(row), row)).with_output_types(
+          Tuple[DateIpKey, Row]))
 
-  if len(all_received) == 0:
-    yield dns_roundtrip
-    return
+  grouped_metadata_and_received_ips = (({
+      BEAM_COGROUP_A_SIDE: tags_keyed_by_ip_and_date,
```
fortuna, line 501: Descriptive names

pipeline/metadata/test_beam_metadata.py
```diff
@@ -31,8 +31,8 @@ def test_merge_metadata_with_rows(self) -> None:
         'date': '2020-01-01',
     }]
     value: Dict[str, List[Row]] = {
-        beam_metadata.IP_METADATA_PCOLLECTION_NAME: [ip_metadata],
-        beam_metadata.ROWS_PCOLLECION_NAME: rows
+        beam_metadata.BEAM_COGROUP_A_SIDE: [ip_metadata],
```
fortuna, line 34: Descriptive names

fortuna (commented):
Looks good. I just think we can use more descriptive key names in the CoGroupByKey

pipeline/beam_tables.py
ohnorobo: Done. I added a third global name

```
RECEIVED_IPS_PCOLLECTION_NAME = 'received_ips'
```

I like how this makes it clearer how the triple-cogroup is working, since it's clearer which if the three parts we're joining when.

The only issue is that we're currently using `merge_metadata_with_rows` for two if the cogroups, and to make that work with the different names I had to add a kvarg to that function. Not entirely sure I like it but let me know.

pipeline/metadata/satellite.py
ohnorobo: done

pipeline/metadata/test_beam_metadata.py
ohnorobo: done

pipeline/metadata/satellite.py
ohnorobo: Agreed these should definitely not be merged. Unmerging them is complicated and unrelated to the main point of this PR though. I'm currently working in it in a [separate branch](https://github.com/censoredplanet/censoredplanet-analysis/compare/triple-cogroup...partition-tags).

pipeline/metadata/satellite.py
fortuna: Let's punt on this for now.

pipeline/metadata/satellite.py
```diff
@@ -375,86 +356,191 @@ def _flat_rows_controls(key: Any, value: Row) -> Iterator[Tuple[Row, int]]:  # p
...
+      received_ips_with_roundtrip_ip_and_date |
+      'received ip: key by ips and dates' >>
+      beam.Map(lambda row: (make_date_ip_key(row), row)).with_output_types(
+          Tuple[DateIpKey, Row]))
 
-  if len(all_received) == 0:
-    yield dns_roundtrip
-    return
+  grouped_metadata_and_received_ips = (({
+      IP_METADATA_PCOLLECTION_NAME: tags_keyed_by_ip_and_date,
```
fortuna, line 504: Let's change the language here to match the language of merge_metadata_with_rows, and remove that extra param.

fortuna (approved):
LGTM, pending that name change

pipeline/metadata/satellite.py
ohnorobo: done

PR #127, opened 2022-03-02
Partition satellite tags

ohnorobo:
Separate out answer and resolver tags as early as possible in the satellite pipeline into their own pcollections. This lets us get rid of the very confusing `_read_satellite_tags` function in favor of separate `_read_satellite_resolver_tags` and `_read_satellite_answer_tags`. 

This does not yet go as far as removing `partition_satellite_input` entirely, since that would require touching code in `beam_tables.py`. But it does simplify it a lot for future refactoring.


pipeline/metadata/satellite.py
```diff
@@ -225,14 +236,9 @@ def _add_vantage_point_tags(
...
-      tags | 'filter out received ip tags' >>
-      beam.Filter(_is_vantage_point_tag).with_output_types(Row))
-
   # PCollection[Tuple[DateIpKey,Row]]
   ips_with_metadata = (
-      vp_tags | 'tags key by ips and dates' >>
+      tags | 'tags key by ips and dates' >>
       beam.Map(lambda row: (make_date_ip_key(row), row)) |
       'combine duplicate tags' >>
       beam.CombinePerKey(_merge_dicts).with_output_types(Tuple[DateIpKey, Row]))
```
fortuna, line 244: Do you need this step? Why not pass `tags` directly to the CoGroupByKey below? That will already group the tags by key. You can remove this step by adjusting grouped_metadata_and_rows.
This will remove an entire shuffle+merge+serialization step.
It's fine if you prefer to do that in a separate PR.

As a general rule of thumb, delay merging as much as possible. Dealing with partitioned data is more efficient than dealing with merged data.

pipeline/metadata/flatten_satellite.py
```diff
@@ -38,20 +38,34 @@
...
+# Files containing metadata for satellite receives answer ips
+SATELLITE_ANSWER_TAG_FILES = [
+    SATELLITE_TAGGED_ANSWERS_FILE, SATELLITE_TAGGED_RESPONSES
+]
+
+# Files containing satellite ip metadata
+SATELLITE_TAG_FILES = SATELLITE_RESOLVER_TAG_FILES + SATELLITE_ANSWER_TAG_FILES
+
+# Files containing satellite blockpage information
+SATELLITE_BLOCKPAGE_FILES = [SATELLITE_BLOCKPAGES_FILE]
```
fortuna, line 55: This is one file. No need for the `_FILES` constant.

pipeline/metadata/satellite.py
```diff
@@ -612,28 +623,15 @@ def partition_satellite_input(
...
     return 1
-  if filename in [
-      flatten_satellite.SATELLITE_INTERFERENCE_FILE,
-      flatten_satellite.SATELLITE_INTERFERENCE_ERR_FILE,
-      flatten_satellite.SATELLITE_ANSWERS_CONTROL_FILE,
-      flatten_satellite.SATELLITE_RESPONSES_CONTROL_FILE,
-      flatten_satellite.SATELLITE_RESULTS_FILE,
-      flatten_satellite.SATELLITE_ANSWERS_ERR_FILE
-  ]:
+  if filename in flatten_satellite.SATELLITE_BLOCKPAGE_FILES:
```
fortuna, line 631: Make it clearer that there's only one file:
```suggestion
  if filename == flatten_satellite.SATELLITE_BLOCKPAGE_FILE:
```

In fact, I preferred when the list was inline here, to it was easier to tell how much work needed to be done for each partition to make them separate inputs.

pipeline/metadata/satellite.py
```diff
@@ -735,18 +733,19 @@ def process_satellite_lines(
     post_processed_satellite: rows of satellite scan data
     blockpage_rows: rows of blockpage data
   """
-  # PCollection[Tuple[filename,line]] x3
-  tags, blockpages, lines = lines | beam.Partition(
+  # PCollection[Tuple[filename,line]] x4
+  answer_lines, resolver_lines, blockpage_lines, row_lines = lines | beam.Partition(
```
fortuna, line 737: Can you remove the blockpages from the merged input and create a separate pcollection from its own filename instead. It should be straightforward in the caller of `process_satellite_lines`

pipeline/metadata/satellite.py
ohnorobo: This is combining metadata which have the same ip and date, but separate tags. In this case different resolver tags are stored in separate files, so we could have a situation like

tag1 {"date":"2022-01-01", "ip":"1.2.3.4", "name": "dns-resolver.com"}
tag2 {"date":"2022-01-01", "ip":"1.2.3.4", "country_code": "US"}

This merges them into a single tag. `merge_metadata_with_rows` currently doesn't know how to deal with cases where we have more than one dict of metadata on the right side of the join.

pipeline/metadata/satellite.py
ohnorobo: It's not as visible in the diff view. But the version of this function with the lists factored out instead of inlined is really much nicer to read.

A5Ji3MiGKL4wHPH

Probably the file lists should be in satellite.py instead of flatten_satellite.py though. I have a future refactoring planned where the filename-conditioning in flatten_satelllite.py also gets pulled up the pipeline to partition much earlier. That should let all the filenames come live here.

pipeline/metadata/satellite.py
ohnorobo: The caller for `process_satellite_lines` is [here](https://github.com/censoredplanet/censoredplanet-analysis/blob/abdac6e698ebbc820d531ac93aabfdcfcf682a84/pipeline/beam_tables.py#L614) in `beam_tables.run_beam_pipeline`. That method has some mixed satellite and hyperquack processing code that needs to be untangled, but I really don't want to add any conditionals on satellite filenames in there. It's also difficult to do without making changes to `_read_scan_text` that would affect hyperquack.

pipeline/metadata/flatten_satellite.py
ohnorobo: removed

pipeline/metadata/satellite.py
ohnorobo: removed the single-element blockpage file list

pipeline/metadata/satellite.py
fortuna: `merge_metadata_with_rows` will get a list under `IP_METADATA_PCOLLECTION_NAME` where each entry will be one of the metadata objects. Just iterate over the list and add the metadata to your target row. There's no need to pre-merge the metadata objects. This also prevents the need for a type for the merged annotations.

pipeline/metadata/satellite.py
fortuna: Ok for now. But Satellite and Hyperquack should also be disentangled.

pipeline/metadata/satellite.py
ohnorobo: This makes me nervous because `merge_metadata_with_rows` is shared code with hyperquack, but okay. Done.

fortuna (approved):

PR #128, opened 2022-03-03
Fix bug where we were writing error fields to the wrong rows

ohnorobo:
Fixes [this bug](https://censoredplanet.slack.com/archives/C01DK5FF4KD/p1646326266983079?thread_ts=1646258098.078749&cid=C01DK5FF4KD) where we were accidentally carrying error fields into the wrong rows. It was caused by reusing a row object incorrectly without making a copy at the right time.

I've also added some functional changes to what we're outputting (not married to any of these):
- only including the `average_confidence`/`matches_confidence` in rows where we have received ips. These are at the top level of the measurement object, but they really only correspond to roundtrips where we get answers.
- changing `success` to just reflect whether we got a 0 rcode. Previously we were using the `connect_error` field to fill this, but that's at the measurement level, whereas `success` is supposed to reflect success at the roundtrip level. We're not using `success` directly in the dashboard in favor of doing our own analysis, so this doesn't make a huge difference. But it's a bit more consistent with the semantics of the field.


pipeline/metadata/flatten_satellite.py
```diff
@@ -388,53 +388,59 @@ def _process_satellite_v2p2(self, row: Row,
     """
     responses = responses_entry.get('response', [])
     row['controls_failed'] = not responses_entry['passed_liveness']
```
fortuna, line 390: Maybe call this `base_row` to be extra explicit?

fortuna (approved):

pipeline/metadata/flatten_satellite.py
ohnorobo: done

PR #131, opened 2022-03-07
fix post processing memory bug

ohnorobo:
I suspect the memory issue here was that we were creating very large lists of rows when co-grouping by (date, domain) and then trying to pass them all as a list to a single function, killing the worker.

Before partitioning `is_control` rows  into the control collection the largest shard here was ('2021-09-29', 'a.root-servers.net') with 9,795,859 rows. After removing these rows the largest shard is ('2021-03-31', 'www.hud.gov') with 93,927 rows.

Ran a a [test backfill](https://console.cloud.google.com/dataflow/jobs/us-central1/2022-03-07_10_55_08-11550896400333842681?project=firehook-censoredplanet) which succeeded.

I'm also still not entirely clear if making this change will still produce outputs with the right meanings for the confidence/excluded fields. @avirkud could you confirm that this matches your intent with the code?

### DIFF
--- a/pipeline/metadata/satellite.py
+++ b/pipeline/metadata/satellite.py
@@ -341,8 +341,7 @@ def _key_by_date_domain(row: Row) -> Tuple[DateDomainKey, Row]:
 def _partition_test_and_control(row_tuple: Tuple[DateDomainKey, Row],
                                 _: int) -> int:
   row: Row = row_tuple[1]
-  # 'anomaly' is None for control measurements
-  if row['is_control_ip'] or row['anomaly'] is None:
+  if row['is_control_ip'] or row['is_control']:
     return 1
   return 0
 
@@ -781,11 +780,10 @@ def process_satellite_lines(
   tagged_satellite = process_satellite_with_tags(row_lines, answer_lines,
                                                  resolver_lines)
 
-  # TODO turn back on once memory problems are fixed
   # PCollection[Row]
-  # post_processed_satellite = post_processing_satellite(tagged_satellite)
+  post_processed_satellite = post_processing_satellite(tagged_satellite)
 
   # PCollection[Row]
   blockpage_rows = process_satellite_blockpages(blockpage_lines)
 
-  return tagged_satellite, blockpage_rows
+  return post_processed_satellite, blockpage_rows
--- a/pipeline/metadata/satellite.py
+++ b/pipeline/metadata/satellite.py
@@ -338,12 +338,22 @@ def _key_by_date_domain(row: Row) -> Tuple[DateDomainKey, Row]:
   return ((row['date'], row['domain']), row)
 
 
-def _partition_test_and_control(row_tuple: Tuple[DateDomainKey, Row],
-                                _: int) -> int:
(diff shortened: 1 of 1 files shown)


pipeline/metadata/satellite.py
```diff
@@ -341,8 +341,7 @@ def _key_by_date_domain(row: Row) -> Tuple[DateDomainKey, Row]:
 def _partition_test_and_control(row_tuple: Tuple[DateDomainKey, Row],
                                 _: int) -> int:
   row: Row = row_tuple[1]
-  # 'anomaly' is None for control measurements
```
fortuna, line 344: Maybe add a todo to make control a separate pcollection.

fortuna (approved):

pipeline/metadata/satellite.py
```diff
@@ -341,8 +341,7 @@ def _key_by_date_domain(row: Row) -> Tuple[DateDomainKey, Row]:
 def _partition_test_and_control(row_tuple: Tuple[DateDomainKey, Row],
```
avirkud, line 341: The original intent of this function was to separate measurements to control resolvers and measurements to test resolvers. (`row['is_control_ip']`)

avirkud: I would maybe change the name to `_partition_test_and_control_resolvers`

pipeline/metadata/satellite.py
```diff
@@ -341,8 +341,7 @@ def _key_by_date_domain(row: Row) -> Tuple[DateDomainKey, Row]:
...
                                 _: int) -> int:
   row: Row = row_tuple[1]
-  # 'anomaly' is None for control measurements
-  if row['is_control_ip'] or row['anomaly'] is None:
+  if row['is_control_ip'] or row['is_control']:
```
avirkud, line 344: The measurements for control domains (`row['is_control']`) should not be in this pcollection. These could be partitioned in a separate function if they are causing the memory bug.

avirkud: I think you could first partition the rows based on the domain (test domain vs. control domain), then run the existing steps on the measurements for test domains (partition test vs. control resolver, etc.)

avirkud (changes requested):

pipeline/metadata/satellite.py
ohnorobo: I'm now partitioning this into three pieces (domain_controls, ip_controls, and tests) and added a doc comment to make it clearer what the intent is.

pipeline/metadata/satellite.py
ohnorobo: I've split this into three pieces. controls that are both domain+ip controls are going into the domain controls pcollection.

pipeline/metadata/satellite.py
ohnorobo: I think we actually don't want to do this. Controls and non-control tests are the same format/schema (by design). And we want to do the same operations on them (adding resolver and answer metadata). The only thing we could buy by splitting them out earlier (I'm assuming you mean in the initial filanema partitioning function) is having to duplicate all those steps twice in our pipeline graph. Currently by only partitioning them just before we need them we're able to keep the total graph simpler.

PR #132, opened 2022-03-08
separate out blockpage processing

ohnorobo:
Split out the blockpage processing to not overlap with the code for the test/control processing. This requires giving the blockpage flattening its own DoFn class.

I initially thought I was going to also split out the test/control measurements, but on closer inspection these will actually remain the same type, just with different values in the `is_control`/`is_ip_control` fields. They're also deeply intertwined in the satellite data, sometimes stored in different files, but sometimes controls are stored inline with the test measurements.

The second commit here is adding a fake BlockpageRow type (actually just Row underneath, but distinct for documentation purposes.) Maybe that should be split off into its own PR which does the same for resolver/answer tags?

### DIFF
--- a/pipeline/metadata/flatten_satellite.py
+++ b/pipeline/metadata/flatten_satellite.py
@@ -19,6 +19,9 @@ from pipeline.metadata.domain_categories import DomainCategoryMatcher
 # Type definition for input responses, TODO make actual type stricter
 ResponsesEntry = Any
 
+# Rows that contain blockpages
+BlockpageRow = Row
+
 SATELLITE_TAGS = {'ip', 'http', 'asnum', 'asname', 'cert'}
 SATELLITE_V2_1_START_DATE = datetime.date(2021, 3, 1)
 SATELLITE_V2_2_START_DATE = datetime.date(2021, 6, 24)
@@ -509,7 +512,7 @@ class FlattenBlockpages(beam.DoFn):
     self.blockpage_matcher = BlockpageMatcher()
     #pylint: enable=attribute-defined-outside-init
 
-  def process(self, element: Tuple[str, str]) -> Iterator[Row]:
+  def process(self, element: Tuple[str, str]) -> Iterator[BlockpageRow]:
     (filename, line) = element
 
     try:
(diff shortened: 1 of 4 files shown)


pipeline/metadata/flatten_satellite.py
```diff
@@ -565,3 +502,63 @@ def _process_satellite_v2_control(
...
+    }
+
+    http = {
+        'https': False,
+    }
+    http.update(row)
+    received_fields = flatten_base.parse_received_data(
+        self.blockpage_matcher, blockpage_entry.get('http', ''), True)
+    http.update(received_fields)
+    yield http
```
fortuna, line 555: Why do you need to emit the http and the https results separately? In the end you will be joining both into the same measurement again, no?

fortuna (changes requested):

pipeline/metadata/flatten_satellite.py
```diff
@@ -565,3 +502,63 @@ def _process_satellite_v2_control(
...
+    try:
+      scan = json.loads(line)
+    except json.decoder.JSONDecodeError as e:
+      logging.warning('JSONDecodeError: %s\nFilename: %s\n%s\n', e, filename,
+                      line)
+      return
+
+    yield from self._process_satellite_blockpages(scan, filename)
+
+  def _process_satellite_blockpages(self, blockpage_entry: ResponsesEntry,
```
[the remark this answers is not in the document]
ohnorobo: It's called that per your request in [PR116](https://github.com/censoredplanet/censoredplanet-analysis/pull/116) [commit](https://github.com/censoredplanet/censoredplanet-analysis/commit/cdfd926f763c4784d1a44b42206ea46d2ca70081) because the response and blockpage data was processed in the same path.

I've split out a `BlockpageEntry` type to match `ResponsesEntry`.

pipeline/metadata/flatten_satellite.py
ohnorobo: They're split because they're fundamentally two independent measurements. We'll probably end up joining both but you could imagine wanting to do different things with them. (Join only the http, or only https, or just look at them independently of the dns.) This way they can be similar to our existing http/s measurement schema, instead of having to have two versions of all their shared fields like `http_received_headers` + `https_received_headers`

fortuna: This is fine with me, but you are essentially creating two new types that are unnecessary (one for HTTP, one for HTTPS).

I believe the pipeline would be more efficient and simpler if the entries from the metadata files were available in their original formats in a big join, so you can extract what you need to create the final output without creating intermediate types. (that's the suggestion in the doc I shared)

fortuna (approved):

PR #133, opened 2022-03-16
strict row types

ohnorobo:
Convert from the `Row = Dict[str, Any]` type we were passing around everywhere to strictly defined dataclasses with fields. Pure refactoring, (except I may have fixed a bug where were were failing to add `name` metadata correctly in some cases.) 

But we were using the `Row` type extensively so this touches almost everything.

`Row` has now been disambiguated into the following types, and it's much easier to understand what we're passing around when.
- `SatelliteAnswer` / `SatelliteAnswerMetadata`
- `IpMetadata`
- `Row`, with child classes `HyperquackRow` and `SatelliteRow`
- `BlockpageRow`
- `ReceivedHttps` (for the parsed content of an HTTP/S request, shared between `HyperquackRow` and `BlockpageRow`)

TESTED: passed e2e test and e2e test with `runtime_type_check=True`.

I'm somewhat worried about the memory implication of passing around dataclasses instead of dicts. Ran a [test job](https://pantheon.corp.google.com/dataflow/jobs/us-central1/2022-03-16_13_47_39-3244885837846865363?project=firehook-censoredplanet) to check, which took 16 hours to complete instead of the usual 9.


pipeline/metadata/schema.py
```diff
@@ -0,0 +1,299 @@
...
+  # Keys
+  ip: str
+  # Metadata
+  asnum: Optional[int] = None
+  asname: Optional[str] = None
+  http: Optional[str] = None
+  cert: Optional[str] = None
+  matches_control: Optional[str] = None
+
+  def add_tag_to_answer(self, tag: SatelliteAnswerMetadata) -> None:
```
fortuna, line 25: Make this a standalone method. No need to be part of the class. It makes the decoupling clearer.

fortuna: Should it be `_tags_`?

pipeline/metadata/schema.py
```diff
@@ -0,0 +1,299 @@
...
+    if tag.http is not None:
+      self.http = tag.http
+    if tag.cert is not None:
+      self.cert = tag.cert
+    if tag.matches_control is not None:
+      self.matches_control = tag.matches_control
+
+
+@dataclass
+class SatelliteAnswerMetadata(SatelliteAnswer):
```
fortuna, line 40: Why don't you use composition instead of inheritance?
It makes moving things around a lot easier (set a field instead of many) and it's easier to understand.

pipeline/metadata/schema.py
```diff
@@ -0,0 +1,299 @@
...
+
+@dataclass
+class IpMetadata():
+  """Class for keeping track of ip metadata"""
+  # Keys
+  ip: str
+  date: str
+  # Metadata
+  netblock: Optional[str] = None
+  asn: Optional[int] = None
```
fortuna, line 59: It would be nice to see AS as a separate type.

pipeline/metadata/schema.py
```diff
@@ -0,0 +1,299 @@
...
+  as_name: Optional[str] = None
+  as_full_name: Optional[str] = None
+  as_class: Optional[str] = None
+  country: Optional[str] = None
+  organization: Optional[str] = None
+  # Satellite Metadata
+  name: Optional[str] = None
```
fortuna, line 66: Maybe rename to `hostname`

pipeline/metadata/schema.py
```diff
@@ -0,0 +1,299 @@
...
+class ReceivedHttps:
+  """Class for the parsed content of a received HTTP/S request
+
+  These are both passed around independantly, and as part of
+  HyperquackRow/BlockpageRow objects
+  """
+  blockpage: Optional[bool] = None
+  page_signature: Optional[str] = None
+
+  received_status: Optional[str] = None
```
fortuna, line 79: Do we need the `received_` prefix?

pipeline/metadata/schema.py
```diff
@@ -0,0 +1,299 @@
...
+
+
+@dataclass
+class ReceivedHttps:
+  """Class for the parsed content of a received HTTP/S request
+
+  These are both passed around independantly, and as part of
+  HyperquackRow/BlockpageRow objects
+  """
+  blockpage: Optional[bool] = None
```
fortuna, line 76: What is this boolean?
Maybe call it `is_known_blockpage`?

pipeline/metadata/flatten_hyperquack.py
```diff
@@ -113,35 +114,35 @@ def _process_hyperquack_v1(self, filename: str, scan: Any,
...
+          controls_failed=scan['FailSanity'],
+          measurement_id=random_measurement_id,
+          source=flatten_base.source_from_filename(filename))
 
       if 'Received' in result:
         received = result.get('Received', '')
         received_fields = flatten_base.parse_received_data(
             self.blockpage_matcher, received, scan['Blocked'])
-        row.update(received_fields)
+        row.update_received(received_fields)
```
fortuna, line 136: You only need this update because you are using inheritance instead of composition.
Why don't you have a `received` field that contains the entire object?
```suggestion
        row.received = received_fields
```

pipeline/metadata/flatten_hyperquack.py
```diff
@@ -159,29 +160,29 @@ def _process_hyperquack_v2(self, filename: str, scan: Any,
...
+          measurement_id=random_measurement_id,
+          source=flatten_base.source_from_filename(filename),
+      )
 
       if 'response' in response:
         received = response.get('response', '')
         received_fields = flatten_base.parse_received_data(
             self.blockpage_matcher, received, scan['anomaly'])
-        row.update(received_fields)
+        row.update_received(received_fields)
```
fortuna, line 183: Same. Prefer composition.

pipeline/metadata/ip_metadata_chooser.py
```diff
@@ -4,45 +4,49 @@
...
-          'netblock': netblock,
-          'asn': asn,
-          'as_name': as_name,
-          'as_full_name': as_full_name,
-          'as_class': as_type,
-          'country': country,
-      }
+      metadata_values = IpMetadata(
+          ip,
+          self.date.isoformat(),
```
fortuna, line 34: Prefer to work with date objects. Call that only where you need to serialize. Beam will do that for you.

Change the `DateIpKey` to use a date object.

pipeline/metadata/schema.py
```diff
@@ -0,0 +1,299 @@
...
+  page_signature: Optional[str] = None
+
+  received_status: Optional[str] = None
+  received_body: Optional[str] = None
+  received_tls_version: Optional[int] = None
+  received_tls_cipher_suite: Optional[int] = None
+  received_tls_cert: Optional[str] = None
+  received_headers: List[str] = dataclasses.field(default_factory=list)
+
+  def update_received(self, new: ReceivedHttps) -> None:
```
fortuna, line 86: Prefer composition over inheritance.
Also, move this to a standalone function.

pipeline/metadata/schema.py
```diff
@@ -0,0 +1,299 @@
...
+      self.received_tls_cipher_suite = new.received_tls_cipher_suite
+    if new.received_tls_cert is not None:
+      self.received_tls_cert = new.received_tls_cert
+    if new.received_headers != []:
+      self.received_headers = new.received_headers
+
+
+# All or part of a scan row to be written to bigquery
+@dataclass
+class Row:  # Corrosponds to BASE_BIGQUERY_SCHEMA
```
fortuna, line 112:
```suggestion
class Row:  # Corresponds to BASE_BIGQUERY_SCHEMA
```

fortuna, line 112: Maybe call it `BigqueryRow` to make it clearer this is the output row (not the input).

pipeline/beam_tables.py
```diff
@@ -609,24 +534,24 @@ def run_beam_pipeline(self, scan_type: str, incremental_load: bool,
...
         self._write_to_bigquery(
             satellite.SCAN_TYPE_BLOCKPAGE, blockpage_rows,
             satellite.get_blockpage_table_name(table_name, scan_type),
             incremental_load)
       else:  # Hyperquack scans
-        # PCollection[Row]
+        # PCollection[HyperquackRow]
         rows = (
             lines | 'flatten json' >> beam.ParDo(
                 flatten.FlattenMeasurement()).with_output_types(Row))
```
fortuna, line 552: Do you need this?
```suggestion
                flatten.FlattenMeasurement()).with_output_types(HyperquackRow))
```

fortuna (changes requested):
I believe things will be a lot less painful if you prefer composition over inheritance.

pipeline/metadata/schema.py
ohnorobo: The main thing I'm trying to achieve here is that the dataclasses being passed around match the ultimate structure we want to write to bigquery, (which is how it currently works with the dicts) so that when we get to the point of [writing the rows](https://github.com/censoredplanet/censoredplanet-analysis/blob/3cf6ac1ef0865bd2ccd616f1861b3701c6ef395f/pipeline/beam_tables.py#L466) we can just call the native `dataclass.asdict` and write them without having to add a translation layer that knows how to map the composed structure into a flat row.

pipeline/metadata/schema.py
ohnorobo: Agreed, but this schema has to match the schema in bigquery, which should change in [PR119](https://github.com/censoredplanet/censoredplanet-analysis/pull/119), not here.

pipeline/metadata/schema.py
ohnorobo: This needs to match the bigquery schema

pipeline/metadata/schema.py
ohnorobo: This is whether the returned match is a known blockpage (True=TP, False=FP, None=no match)

These names can't change here since they need to match the existing bq schema.

pipeline/metadata/beam_metadata.py
```diff
@@ -13,43 +14,141 @@
...
   else:
     ip_metadatas = []
-  rows = value[ROWS_PCOLLECION_NAME]
+  rows: List[Row] = value[ROWS_PCOLLECION_NAME]  # type: ignore
 
   for row in rows:
-    new_row: Row = {}
-    new_row.update(row)
+    new_row = deepcopy(row)
     for ip_metadata in ip_metadatas:
```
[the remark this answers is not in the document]
ohnorobo: This used to only be one, but then we removed the [metadata deduping here](https://github.com/censoredplanet/censoredplanet-analysis/pull/127#discussion_r819755803) so now there can be multiple.

pipeline/beam_tables.py
ohnorobo: Yes, the hyperquack and satellite flows are different. I'm not a big fan of how they're intertwined here, but that's not that I'm trying to fix in this change.

pipeline/metadata/ip_metadata_chooser.py
ohnorobo: Grouping on keys that are not strings is not fully automatic in beam. There's [some trickiness](https://stackoverflow.com/questions/69409693/how-do-i-use-a-datetime-date-value-in-apache-beam-groupby) at least.

Can I attempt this in a seperate change since it's independent of what I'm doing in this PR?

pipeline/metadata/schema.py
ohnorobo: As in a composed subtype with AS information, or as in a `type asn = int` type?

pipeline/metadata/schema.py
ohnorobo: made standalone, and changed name to `tags`

pipeline/metadata/schema.py

pipeline/metadata/schema.py
ohnorobo: moved to standalone

pipeline/metadata/schema.py
fortuna: The serialization code is something you would write once. And you are writing something equivalent anyway in your dict update-like functions, with the difference that your logic is spread throughout different places.

It's important to decouple the structured used in code from the output format that has to stay backwards-compatible. Using the output format internally prevents us from being able to make changes.

My recommendation is to use composition and have a clear translation step at the end. It will free yourself to keep changing the internal format without breaking compatibility, and the serialization step will work as runnable documentation of how the output is generated.

pipeline/metadata/schema.py
fortuna: A composable subtype, with fields like number, name, etc (I had that in my SQL query)

pipeline/metadata/schema.py
fortuna: This is a clear example of how the lack of a serialization layer prevents us from making changes

pipeline/metadata/schema.py
fortuna: Again

pipeline/metadata/ip_metadata_chooser.py
fortuna: This is not a key here. You could convert to string where you extract the key. But, yeah, we don't need to address that in this PR.

pipeline/metadata/beam_metadata.py
```diff
@@ -2,9 +2,10 @@
 
 from __future__ import absolute_import
 
-from typing import Tuple, Dict, List, Iterator
+from copy import deepcopy
+from typing import Tuple, Dict, List, Iterator, Union, Iterable
 
-from pipeline.metadata.flatten import Row
+from pipeline.metadata.schema import BigqueryRow, SatelliteRow, IpMetadata, SatelliteAnswer, SatelliteAnswerMetadata, add_metadata_to_row, add_tags_to_answer
```
fortuna, line 8: Is the PR ready for review?

I don't see `add_metadata_to_row` anymore, for instance.

ohnorobo:
Switched to composition instead of inheritance as discussed.

The dict-update-like methods for `IpMetadata` and `SatelliteAnswers` ended up being required. For both of them we need a way to merge multiple disparate sources of metadata/tags while ignoring nulls. The version for `ReceivedHttps` data is gone though.

I've added the serialization methods in `schema.py` as `flatten_for_bigquery_*`

pipeline/metadata/schema.py
```diff
@@ -0,0 +1,402 @@
...
+
+  When this metadata is being passed around
+  it needs an additional date field to keep track of when it's valid.
+  In the final SatelliteAnswer object written to bigquery
+  we don't include that field since it's redundant.
+  """
+  date: str = ''
+
+
+def merge_satellite_tags(answer: SatelliteAnswer,
```
fortuna, line 38: Pass two `AnnotatedSatelliteAnswer` objects instead.
You can get the second parameter from the DatedAnnotatedSatelliteAnswer with `dasa.answer`.

fortuna (approved):
Looks good. I'm adding some comments for a potential future PR

PR #134, opened 2022-03-21
add missing DNS country codes

ohnorobo:
Add some country codes that were new in the DNS data and not getting matched


eltsai (approved):
Looks good to me!

PR #135, opened 2022-03-22
make measurement_ids permanent instead of random

ohnorobo:
Small change that's been on my mind.

Currently we generate the measurement ids randomly and they change every time we re-run the pipeline. This instead generates them from a SHA1 hash of the filename + the measurement line, so they'll be stable across backfills. This is to potentially enable some far-future project where users can view full measurements from a permanent url.

I've used the string representation instead of the parsed json or just picking some fields since I think the string representation will be stable and unique when hashed, whereas something like the json might be sensitive to python implementation changes.

The inclusion of the filename is to make sure ids are unique across scans. I don't think scans can really generate duplicate lines (because of the unique date field) but better safe than sorry. Am I missing any other source of potential duplicate ids?

### DIFF
--- a/pipeline/metadata/flatten.py
+++ b/pipeline/metadata/flatten.py
@@ -15,6 +15,9 @@ from pipeline.metadata.domain_categories import DomainCategoryMatcher
 from pipeline.metadata.flatten_satellite import SatelliteFlattener, SATELLITE_PATH_COMPONENT
 from pipeline.metadata.flatten_hyperquack import HyperquackFlattener
 
+# Randomly chosen UUID used as a namespace for generating further UUIDs
+CENSORED_PLANET_NAMESPACE = uuid.UUID('23fd299b-6547-4dca-b630-cb05dec70e0e')
+
 
 class FlattenMeasurement(beam.DoFn):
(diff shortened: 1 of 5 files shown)


pipeline/metadata/flatten.py
```diff
@@ -15,6 +15,9 @@
 from pipeline.metadata.flatten_satellite import SatelliteFlattener, SATELLITE_PATH_COMPONENT
 from pipeline.metadata.flatten_hyperquack import HyperquackFlattener
 
+# Randomly chosen UUID used as a namespace for generating further UUIDs
+CENSORED_PLANET_NAMESPACE = uuid.UUID('23fd299b-6547-4dca-b630-cb05dec70e0e')
```
fortuna, line 19: You could also use something like `uuid.uuid5(uuid.NAMESPACE_DNS, "censoredplanet.org")`

fortuna (approved):

fortuna (approved):

ohnorobo:
Following up on this, the new ids are in the hyperquack tables and I'm not seeing any problems with duplicates.
image

PR #136, opened 2022-03-24
follow up refactoring

ohnorobo:
Some small pieces of follow-up to https://github.com/censoredplanet/censoredplanet-analysis/pull/133

Contains:
- remove `received_` prefixes from ReceivedHttps fields
- split SatelliteAnswer into some subpieces. This will make it easier to add other subpieces in the future.
	- ip field
	- IpMetadata field (for the asn/as name)
	- SatelliteTags field for the http/cert/matches_control tags
- change typedefs of `merge*` functions to be adding the same types together
- use dataclass default_factory to make sure dataclass subfields are instantiated correctly.
- made `_calculate_confidence` somewhat more readable

I went with the names `IpMetadata` / `IpMetadataWithKeys` and `SatelliteAnswer` / `SatelliteAnswerWithKeys` for the versions with extra key fields.


pipeline/metadata/schema.py
```diff
@@ -93,6 +53,49 @@ def merge_ip_metadata(base: IpMetadata, new: IpMetadataWithKeys) -> None:
...
+
+def merge_satellite_answers(base: SatelliteAnswer,
+                            new: SatelliteAnswer) -> None:
+  """Add tag information to a Satellite answer"""
+  merge_satellite_tags(base.tags, new.tags)
+  merge_ip_metadata(base.ip_metadata, new.ip_metadata)
+
+
 @dataclass
 class ReceivedHttps:
```
fortuna, line 100: Optional: This is a weird name. Maybe `HttpsResponse` would be clearer.

pipeline/metadata/schema.py
```diff
@@ -103,12 +106,12 @@ class ReceivedHttps:
   blockpage: Optional[bool] = None
```
fortuna, line 106: Can we rename this to `is_known_blockpage`? This name is confusing.

fortuna (approved):
This is strictly better, thanks. I think there's still room to explore in the format, but perhaps we can do that in a doc, not here.

pipeline/metadata/schema.py
```diff
@@ -93,6 +53,49 @@ def merge_ip_metadata(base: IpMetadata, new: IpMetadataWithKeys) -> None:
...
+  cert: Optional[str] = None
+  matches_control: Optional[str] = None
+
+
+@dataclass
+class SatelliteAnswer():
+  """Class for keeping track of Satellite answer content"""
+  # Keys
+  ip: str
+  tags: SatelliteTags = dataclasses.field(default_factory=SatelliteTags)
```
fortuna, line 69: This "tags" concept is a bit arbitrary and is mixing things we observed (http_body, cert), with things we derive (matches_control)
Can you keep those separate?
I'd just keep them inline here for now, rather than introducing a concept that is not cleaar.

fortuna (changes requested):

pipeline/metadata/schema.py
ohnorobo: https://github.com/censoredplanet/censoredplanet-analysis/pull/137

pipeline/metadata/schema.py

pipeline/metadata/schema.py

fortuna (approved):

PR #138, opened 2022-03-25
inline blockpage responses with satellite answers

ohnorobo:
Inline returned http/s information from testing anomalous ips instead of outputting a separate blockpage table.

Successful run [here](https://console.cloud.google.com/dataflow/jobs/us-central1/2022-03-31_10_17_09-7149755111185919793?project=firehook-censoredplanet)

test dashboard [here](https://datastudio.google.com/c/u/0/reporting/46f457be-2182-43eb-8a2d-db2f8d9b2465).

Performance over the run:
image

Having "adding tags to answers" and "adding blockpages to answers" as two separate operations in series is not the most efficient way to do this. (Both of those operations are repeating some of the same work) But it makes the code very easy to understand, and I'd prefer to get it working and checked in before working on efficiency improvements.

Naively adding the inline data was causing the postprocessing code to fail with memory issues. Presumably grouping rows by (date, domain) caused shards which repeated the same http bodies many times and were too big. I've rewritten the postprocessing to use a side-input instead of a cogroup to avoid this.

I also moved some of the bigquery schema code from `beam_tables.py` to `schema.py`


pipeline/metadata/beam_metadata.py
```diff
@@ -153,3 +170,37 @@ def merge_tagged_answers_with_rows(
       if tags.ip == untagged_answer.ip:
         merge_satellite_answers(untagged_answer, tags)
   return row
+
+
+def merge_blockpages_with_answers(
```
fortuna, line 175: I'd use `add_page_fetch_to_answers` or something like that. Let's move away from `blockpage` as it's not specific to blockpages.

pipeline/metadata/beam_metadata.py
```diff
@@ -153,3 +170,37 @@ def merge_tagged_answers_with_rows(
...
+    key: DomainDateIpKey,  # pylint: disable=unused-argument
+    value: Dict[str, Union[List[SatelliteAnswer],
+                           List[Tuple[str, SatelliteAnswerWithKeys]]]]
+) -> Iterator[Tuple[str, SatelliteAnswer]]:
+  """Add blockpage info to answers."""
```
fortuna, line 180: Maybe:
```suggestion
  """Add page information to answers."""
```

pipeline/metadata/beam_metadata.py
```diff
@@ -153,3 +170,37 @@ def merge_tagged_answers_with_rows(
...
+  answers: List[Tuple[str, SatelliteAnswer]] = value[
+      RECEIVED_IPS_PCOLLECTION_NAME]  # type: ignore
+  blockpages: List[BlockpageRow] = value[
```
fortuna, line 183: Rename `blockpages` and `BlockpageRow`

pipeline/metadata/beam_metadata.py
```diff
@@ -153,3 +170,37 @@ def merge_tagged_answers_with_rows(
...
+      BLOCKPAGE_PCOLLECTION_NAME]  # type: ignore
+
+  if not blockpages:
+    for (roundtrip_id, answer) in answers:
+      yield (roundtrip_id, answer)
+
+  https_blockpages = [
+      blockpage for blockpage in blockpages if blockpage.https is True
```
fortuna, line 191: I'd point out that you are flattening and merging http and https, just to separate them again here.
My recommendation is to keep them both in the same structure. There's no point in separating them. Or have them be separate inputs handled by separate functions.

The red flag here is that HTTP and HTTPS fetches here have different types, but are put in the same collection.

pipeline/metadata/schema.py
```diff
@@ -7,8 +7,31 @@
 from dataclasses import dataclass
 from typing import Optional, List, Dict, Any, Union
 
+from apache_beam.io.gcp.internal.clients import bigquery as beam_bigquery
+
 # pylint: disable=too-many-instance-attributes
 
+SCAN_TYPE_SATELLITE = 'satellite'
+SCAN_TYPE_BLOCKPAGE = 'blockpage'
```
fortuna, line 15: I really want us to move away from the "blockpage" mentality. Also, block pages don't even make sense for the most part for HTTPS. Use "page fetch", "answer_page" or something that reflects it's about the retrieval of the domain page from the answer IP.

pipeline/metadata/schema.py
```diff
@@ -246,13 +260,32 @@ def flatten_for_bigquery_satellite(row: SatelliteRow) -> Dict[str, Any]:
...
-        'asnum': received_answer.ip_metadata.asn,
-        'asname': received_answer.ip_metadata.as_name,
         'http': received_answer.http,
         'cert': received_answer.cert,
         'matches_control': received_answer.matches_control,
+        # Ip Metadata
+        'asnum': received_answer.ip_metadata.asn,
+        'asname': received_answer.ip_metadata.as_name,
+        # HTTP
+        'http_response_is_known_blockpage': http_response.is_known_blockpage,
```
fortuna, line 275: You could use a structure for `http_response` and `https_response` instead. I believe a structure helps understand the schema better.

pipeline/metadata/schema.py
```diff
@@ -246,13 +260,32 @@ def flatten_for_bigquery_satellite(row: SatelliteRow) -> Dict[str, Any]:
...
+        'http_response_page_signature': http_response.page_signature,
+        'http_response_status': http_response.status,
+        'http_response_body': http_response.body,
+        'http_response_headers': http_response.headers,
```
fortuna, line 279: Minor style suggestion: let's order the fields in the order we observe or derive them:
```suggestion
        'http_response_status': http_response.status,
        'http_response_headers': http_response.headers,
        'http_response_body': http_response.body,
        'http_response_page_signature': http_response.page_signature,
        'http_response_is_known_blockpage': http_response.is_known_blockpage,
```

Same for HTTPS

pipeline/metadata/schema.py
```diff
@@ -246,13 +260,32 @@ def flatten_for_bigquery_satellite(row: SatelliteRow) -> Dict[str, Any]:
...
+        # HTTPS
+        'https_response_is_known_blockpage': https_response.is_known_blockpage,
+        'https_response_page_signature': https_response.page_signature,
+        'https_response_status': https_response.status,
+        'https_response_body': https_response.body,
+        'https_response_headers': https_response.headers,
+        'https_response_tls_version': https_response.tls_version,
+        'https_response_tls_cipher_suite': https_response.tls_cipher_suite,
+        'https_response_tls_cert': https_response.tls_cert,
```
fortuna, line 288:
```suggestion
        'https_response_tls_version': https_response.tls_version,
        'https_response_tls_cipher_suite': https_response.tls_cipher_suite,
        'https_response_tls_cert': https_response.tls_cert,
        'https_response_status': https_response.status,
        'https_response_headers': https_response.headers,
        'https_response_body': https_response.body,
        'https_response_page_signature': https_response.page_signature,
        'https_response_is_known_blockpage': https_response.is_known_blockpage,
```

pipeline/metadata/schema.py
fortuna, line 275: Perhaps we separate what is observed (immutable) from what is derived (can be recalculated) by renaming these two fields to something like `http_analysis_page_signature` and `http_analysis_is_known_blockpage`.
Thoughts?

pipeline/metadata/schema.py
```diff
@@ -342,17 +375,36 @@ def _add_schemas(schema_a: Dict[str, Any],
...
+            'repeated',
+            {
+                'ip': ('string', 'nullable'),
+                'asnum': ('integer', 'nullable'),
+                'asname': ('string', 'nullable'),
+                'http': ('string', 'nullable'),
+                'cert': ('string', 'nullable'),
+                'matches_control': ('string', 'nullable'),
+                # HTTP
+                'http_response_is_known_blockpage': ('boolean', 'nullable'),
```
fortuna, line 393: Field ordering.  Could benefit from structure.

pipeline/metadata/beam_metadata.py
ohnorobo: done

pipeline/metadata/beam_metadata.py
ohnorobo: done

pipeline/metadata/beam_metadata.py
ohnorobo: done

pipeline/metadata/schema.py
ohnorobo: I don't think it's really a mentality so much as that the file we're reading these from is called `blockpages.json`,  so it made some sense to keep using that name as we passed it around.

But done

pipeline/metadata/schema.py
ohnorobo: renamed

pipeline/metadata/beam_metadata.py
ohnorobo: They're the same type (in the same way that test rows and control rows are the same type, but have different `is_control` fields). But they could have something like a parent type which holds two of them.

But it did make more sense to separate them when we were writing them to their own table. Now with the inlining it's less useful.

pipeline/metadata/schema.py
ohnorobo: commented above

pipeline/metadata/schema.py
```diff
@@ -246,19 +260,38 @@ def flatten_for_bigquery_satellite(row: SatelliteRow) -> Dict[str, Any]:
...
         'cert': received_answer.cert,
         'matches_control': received_answer.matches_control,
+        # Ip Metadata
+        'asnum': received_answer.ip_metadata.asn,
+        'asname': received_answer.ip_metadata.as_name,
+        # HTTP
+        'http_response_status': http_response.status,
+        'http_response_headers': http_response.headers,
+        'http_response_body': http_response.body,
+        'http_response_page_signature': http_response.page_signature,
```
fortuna, line 278: Rename to `http(s)_analysis_`?

fortuna (approved):

pipeline/metadata/schema.py

PR #139, opened 2022-04-01
fix dns page response error parsing

ohnorobo:
We were accidentally parsing the DNS HTTP/S response page errors into the `http_received_status` and `https_received_status` fields. This parses them into their own `http_error` and `https_error` fields.


eltsai (approved):
Looks good to me! Thanks for tagging me in this PR (so I got to read the pipeline and understood that we are actually flattening the responses for a `(ip, domain)` pair into 2 lines).

PR #148, opened 2022-04-11
Make matches_control a struct

ohnorobo:
For the [schema name changes](https://docs.google.com/document/d/1BMohMUiqJzYnTR8fdEqb1Ang5r1oW_aEHdsxQraQPR8/edit#).

Turn `matches_control`, which was a string field like `ip http asname` into a structured field of booleans like

```
{
 ip = True,
 http = True,
 cert = False,
 asnum = False,
 asname = True
}
```

Both internally in the code and externally in the bq schema.

### DIFF
--- a/pipeline/metadata/flatten_satellite.py
+++ b/pipeline/metadata/flatten_satellite.py
@@ -116,12 +116,17 @@ def _make_matches_control(tags: List[str]) -> MatchesControl:
   Returns:
     string like "ip http cert"
   """
-  matches_control = MatchesControl(
-      ip='ip' in tags,
-      http='http' in tags,
-      cert='cert' in tags,
-      asnum='asnum' in tags,
-      asname='asname' in tags)
+  matches_control = MatchesControl()
+  if 'ip' in tags:
+    matches_control.ip = True
+  if 'http' in tags:
+    matches_control.http = True
+  if 'cert' in tags:
+    matches_control.cert = True
+  if 'asnum' in tags:
+    matches_control.asnum = True
+  if 'asname' in tags:
+    matches_control.asname = True
   return matches_control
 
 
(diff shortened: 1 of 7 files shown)


pipeline/metadata/flatten_satellite.py
```diff
@@ -117,7 +116,13 @@ def _append_tags(tags: List[str]) -> str:
   Returns:
     string like "ip http cert"
   """
-  return ' '.join([tag for tag in tags if tag in SATELLITE_TAGS])
+  matches_control = MatchesControl(
+      ip='ip' in tags,
```
fortuna, line 120: This is not the right semantics. You should only set to false when the test actually happened.
CP is not checking IPs or AS when the AS information is missing, for instance.

For now, let's output True if present, and NULL otherwise.
If we add the control test to the pipeline, then we can correctly set the fields.

pipeline/metadata/schema.py
```diff
@@ -82,14 +82,24 @@ def merge_ip_metadata(base: IpMetadata, new: IpMetadata) -> None:
     base.name = new.name
 
 
+@dataclass
+class MatchesControl():
+  """Class to keep track of which answer fields matched the control."""
+  ip: bool = False
```
fortuna, line 88: Make them None by default.

pipeline/metadata/flatten_satellite.py
fortuna: BTW, one easy way to do that is to add the trusted answers as a domain tag.

fortuna (changes requested):

pipeline/metadata/schema.py

fortuna (approved):
Thanks for the change. This should really change in the source data, since that's where they really know when the test happened.

PR #151, opened 2022-04-15
move outcome classification to python

ohnorobo:
Move the hyperquack outcome classification code into python from sql.

This is almost a pure refactoring. (no changes to the actual outcome strings) We're just moving the calculation to python so the field can be output in the base table as well.

In a follow up we may want to make some changes to the strings (like splitting the "stage/outcome" fields into different strings.)

Ran test backfills to compare outcomes against: [echo](https://pantheon.corp.google.com/dataflow/jobs/us-west1/2022-04-20_14_33_02-6209285574164862714?project=firehook-censoredplanet) [discard](https://pantheon.corp.google.com/dataflow/jobs/us-west2/2022-04-20_14_35_08-215118299917237301?project=firehook-censoredplanet) [http](https://pantheon.corp.google.com/dataflow/jobs/us-east4/2022-04-20_14_33_00-5276552522047621665?project=firehook-censoredplanet) [https](https://pantheon.corp.google.com/dataflow/jobs/us-east1/2022-04-20_14_33_03-7809081176646427166?project=firehook-censoredplanet)

image

Tested the pipeline-generated outcomes against the old sql generated ones.
image

I found one case where we weren't matching related to a typo. [Fixed that](https://github.com/censoredplanet/censoredplanet-analysis/pull/151/commits/c569bf5eaaa8235b2e3ffd2385967aba2b20a57c).


pipeline/metadata/flatten_hyperquack.py
```diff
@@ -135,6 +153,11 @@ def _process_hyperquack_v1(self, filename: str, scan: Any,
       if 'Error' in result:
         row.error = result['Error']
 
+      row.outcome = classify_hyperquack_outcome(
```
fortuna, line 156: I worry that we are losing the ability to easily regenerate the outcome.
Perhaps you should make this a separate step, a PTransform that is adds the outcome row. This way we could reuse the same PTranform to update the data, with the existing data as input.

Regardless, I don't think there's significant cost in making this a separate transform, but you will get better metrics and the semantics is clearer. This should run on the output of process_hyperquack. Nesting here makes the processing step harder to understand.

pipeline/metadata/hyperquack_outcome.py
```diff
@@ -0,0 +1,170 @@
...
+    matches_template: did the response match the control template
+    is_known_blockpage: did the page content match a blockpage
+    blockpage_fingerprint: which known blockpage did we match
+    received_headers: HTTP headers
+
+
+  Returns:
+    outcome string like "read/tcp.reset"
+  """
+  if is_known_blockpage and blockpage_fingerprint:
```
fortuna, line 158: Maybe add a TODO to order this by when we see the signals.

pipeline/metadata/hyperquack_outcome.py
```diff
@@ -0,0 +1,170 @@
...
+  """
+  if errors is None:
+    return None
+
+  if ';' in errors:
+    return errors[:errors.index(';')]
+  return errors  # single error
+
+
+def _has_akamai_server_header(received_headers: List[str]) -> bool:
```
fortuna, line 37: I wish the headers were properly structured. I'd highly recommend that change for the data dumps.
Something like `STRUCT(name: "Server", value: "AkamaiGHost")`

Then you can write code like `if (headers.get("Server") in [...])`

pipeline/metadata/hyperquack_outcome.py
```diff
@@ -0,0 +1,170 @@
...
+  # yapf: disable
+  # pylint: disable=multiple-statements
+  # pylint: disable=too-many-return-statements
+  # pylint: disable=too-many-branches
+  # pylint: disable=too-many-statements
+
+  # Success
+  if error is None or error == "": return "expected/match"
+
+  # System failures
```
fortuna, line 63: Move this to a `is_system_failure` method, and do:
```py
if (is_system_failure(error)): return "setup/system_failure"
```

Move it to classify_hyperquack_outcome

pipeline/metadata/hyperquack_outcome.py
```diff
@@ -0,0 +1,170 @@
...
+  if error.endswith("address already in use"): return "setup/system_failure"
+  if error.endswith("protocol error"): return "setup/system_failure"  # ipv4 vs 6 error
+  if error.endswith("protocol not available"): return "setup/system_failure"
+  if error.endswith("too many open files"): return "setup/system_failure"
+
+  # Dial failures
+  if error.endswith("network is unreachable"): return "dial/ip.network_unreachable"
```
fortuna, line 70: Do something like this:
```py
dial_error = get_dial_error(error)
if dial_error: return f"dial/{dial_error}"
```

So the dial logic is encapsulated. You could even imagine outputting a `dial_error` field in the future.

Move it to classify_hyperquack_outcome

pipeline/metadata/hyperquack_outcome.py
```diff
@@ -0,0 +1,170 @@
...
+  if error.endswith("no route to host"): return "dial/ip.host_no_route"
+  if error.endswith("connection refused"): return "dial/tcp.refused"
+  if error.endswith("context deadline exceeded"): return "dial/timeout"
+  if error.endswith("connect: connection timed ou"): return "dial/timeout"
+  if error.startswith("connection reset by peer"): return "dial/tcp.reset"  # no read: or write: prefix in error
+  if error.endswith("connect: connection reset by peer"): return "dial/tcp.reset"
+  if error.endswith("getsockopt: connection reset by peer"): return "dial/tcp.reset"
+
+  # TLS failures
+  if "tls:" in error: return "tls/tls.failed"
```
fortuna, line 80: Same approach:
```py
tls_error = get_tls_error(error)
if tls_error: return f"tls/{tls_error}"
```

Move it to classify_hyperquack_outcome

fortuna: Same for the other steps below.

fortuna (changes requested):
I believe we can better encapsulate each step in the outcome calculation as if we had a field for each step. And move all steps to classify_hyperquack_outcome, so we can read the entire logic in one place.

pipeline/metadata/flatten_hyperquack.py
ohnorobo: Broke off a refactoring I need for this into https://github.com/censoredplanet/censoredplanet-analysis/pull/153

pipeline/metadata/hyperquack_outcome.py
ohnorobo: done

pipeline/metadata/hyperquack_outcome.py
ohnorobo: done

pipeline/metadata/hyperquack_outcome.py
ohnorobo: done

pipeline/metadata/hyperquack_outcome.py
ohnorobo: done

pipeline/metadata/flatten_hyperquack.py
ohnorobo: Done, moved outcome classification to the end of `process_hyperquack_lines`

pipeline/metadata/hyperquack_outcome.py
```diff
@@ -0,0 +1,202 @@
...
+
+  if _has_akamai_server_header(received_headers):
+    return "expected/trusted_host:akamai"
+
+  # Content mismatch for hyperquack v2 which didn't write
+  # content verification failures in the error field from 2021-04-26 to 2021-07-21
+  if not matches_template and (error is None or error == ""):
+    return "content/mismatch"
+
+  return _classify_hyperquack_error(
```
fortuna, line 201: I'd inline `_classify_hyperquack_error` here, so the entire high level logic is read in one place.

fortuna (approved):

pipeline/metadata/hyperquack_outcome.py
ohnorobo: done

PR #152, opened 2022-04-18
Export data to GCS

avirkud:
Adding an option to export the processed data to GCS instead of BigQuery: `--export_gcs`

Test jobs: [1](https://console.cloud.google.com/dataflow/jobs/us-central1/2022-04-14_12_29_44-8898090544698600642?project=firehook-censoredplanet), [2](https://console.cloud.google.com/dataflow/jobs/us-west2/2022-04-18_13_45_33-4502948207900998221?project=firehook-censoredplanet)

Files are written to `gs://{bucket}/{dataset}/{scan_type}/source={source}/country={country}/results.json.gz`,
for example `gs://firehook-test/avirkud/satellite/source=CP_Satellite-2022-01-02-12-00-01/country=AD/results.json.gz`.
- bucket is currently `firehook-test` (TODO: update this)
- dataset is either `base` or `<username>`
- scan_type is one of `echo`, `discard`, `http`, `https`, or `satellite`. We discussed changing `satellite` -> `dns`. I'm leaving this change for a later PR since it would need to be changed in a few places.


pipeline/beam_tables.py
```diff
@@ -182,23 +205,25 @@ def _get_partition_params() -> Dict[str, Any]:
...
 
-def get_job_name(table_name: str, incremental_load: bool) -> str:
+def get_job_name(destination: str, incremental_load: bool) -> str:
   """Creates the job name for the beam pipeline.
 
   Pipelines with the same name cannot run simultaneously.
 
   Args:
-    table_name: a dataset.table name like 'base.scan_echo'
+    table_name: a dataset.table name like 'base.scan_echo' or
```
ramakrishnansr, line 214: More of a nit, but this variable can be renamed.

ramakrishnansr (approved):

ohnorobo:
> A couple of things:
> 
> * Add a translation layer to make sure the GCS schema is not tied to the BigQuery schema. We should be able to change them independently.
> * I'd highly recommend some schema cleanup. Sarah is doing some renaming. Also, some fields like matches control should be a struct and headers field should be split into header name and value for ease of use. (the client shouldn't have to parse anything). There may be other cleanup.

I don't think we want an another separate translation layer that can get out of sync. I think the rule should be that the output json files *must* use the same schema as the output tables, and the table translation layer gets to determine things like field types and names. (Meaning this PR will be dependant on https://github.com/censoredplanet/censoredplanet-analysis/pull/119 for what the final names are.) 
The only additional translation layer I would add is one that is able to remove some fields, since we know we want to do that.

fortuna:
@ohnorobo  The layer to remove fields is the translation layer. To start the schemas can be the same. But we should be able to change the BigQuery schema without breaking consumers of the data dumps. How will we be able to do that without the translation layer?
Also, if we add something to BigQuery, it shouldn't automatically appear on the dump. What goes in the dump should be vetted.
Being out of sync is working as intended.

ohnorobo:
> @ohnorobo The layer to remove fields is the translation layer. To start the schemas can be the same. But we should be able to change the BigQuery schema without breaking consumers of the data dumps. How will we be able to do that without the translation layer? Also, if we add something to BigQuery, it shouldn't automatically appear on the dump. What goes in the dump should be vetted. Being out of sync is working as intended.

I'm worried that we'll end up reinventing the wheel and have two very similar schemas with slightly different names.

My suggestion is that the current `flatten_for_bigquery_*(row) -> Dict:` methods should be changed to just `flatten_to_dict-*` and shared between the json file writing and the bq writing. Then each output can have its own extra `dict_to_bigquery`/`dict_to_json_string` layer that can make arbitrary changes, drop fields, etc, but we'll have less of the schemas drifting away from each other just because of arbitrary naming choices.

pipeline/beam_tables.py
```diff
@@ -59,26 +62,46 @@
...
 
 
-def _get_existing_datasources(table_name: str, project: str) -> List[str]:
-  """Given a table return all sources that contributed to the table.
+def _get_existing_datasources(destination: str, project: str) -> List[str]:
+  """Given a BigQuery or GCS destination return all contributing sources.
 
   Args:
-    table_name: name of a bigquery table like
+    destination: name of a bigquery table like
```
ohnorobo, line 69: I would not overload this field to have two different types of strings. Instead I would break this out into two methods  `_get_existing_bq_datasources` and `_get_existing_file_datasources`

pipeline/beam_tables.py
```diff
@@ -182,23 +205,25 @@ def _get_partition_params() -> Dict[str, Any]:
   return partition_params
 
 
-def get_job_name(table_name: str, incremental_load: bool) -> str:
+def get_job_name(destination: str, incremental_load: bool) -> str:
```
ohnorobo, line 208: same here. I would break this into two methods, maybe with a shared helper.

pipeline/beam_tables.py
```diff
@@ -265,7 +304,7 @@ def _data_to_load(self,
                     gcs: GCSFileSystem,
                     scan_type: str,
                     incremental_load: bool,
-                    table_name: str,
+                    destination: str,
```
ohnorobo, line 307: I would turn this into two optional kv args `table_name` and `file_destination` and switch on whether those are set instead of switching on the content of the string.

pipeline/beam_tables.py
```diff
@@ -425,6 +464,70 @@ def _write_to_bigquery(self, scan_type: str,
...
+        'gs://firehook-test/scans/echo' to write output files to.
+
+    Raises:
+      Exception: if any arguments are invalid.
+    """
+
+    def _get_destination(record: str) -> str:
+      """Returns the hive-format dest folder for a measurement record str."""
+      record_dict = json.loads(record)
+      return f'source={record_dict["source"]}/country={record_dict["country"]}/results'
```
ohnorobo, line 489: @fortuna I'm not convinced the hive format filepaths are a good idea unless we have a strong sense of who would want to use it. It adds extra confusion for people who are just trying to read the files and parse them some other way. We can discuss in the meeting.

pipeline/beam_tables.py
```diff
@@ -425,6 +464,70 @@ def _write_to_bigquery(self, scan_type: str,
...
+        logging.info('%s %s %s %s', destination, shard_index, total_shards,
+                     filename)
+        return filename
+
+      return _inner
+
+    # Pcollection[Dict[str, Any]]
+    dict_rows = (
+        rows | f'dataclass to dicts {scan_type}' >> beam.Map(
+            schema.flatten_for_bigquery).with_output_types(Dict[str, Any]))
```
ohnorobo, line 515: Discussed elsewhere, but we should rename this/add an extra translation layer if we're using it for json.

pipeline/beam_tables.py
```diff
@@ -425,6 +464,70 @@ def _write_to_bigquery(self, scan_type: str,
...
+
+    # PCollection[str]
+    json_rows = (
```
ohnorobo, line 518: If we do the comment above, this and the step above will become a map over a single method which would live in schema.py

pipeline/beam_tables.py
```diff
@@ -453,31 +556,44 @@ def _get_pipeline_options(self, scan_type: str,
     return pipeline_options
 
   def run_beam_pipeline(self, scan_type: str, incremental_load: bool,
-                        job_name: str, table_name: str,
+                        job_name: str, destination: str,
```
ohnorobo, line 559: break this out into two args instead of overloading

pipeline/run_beam_tables.py
```diff
@@ -54,13 +57,17 @@ def run_parallel_pipelines(runner: beam_tables.ScanDataBeamPipelineRunner,
...
     for scan_type in scan_types:
-      table_name = beam_tables.get_table_name(dataset, scan_type,
-                                              beam_tables.BASE_TABLE_NAME)
-      job_name = beam_tables.get_job_name(table_name, incremental_load)
+      destination = beam_tables.get_table_name(dataset, scan_type,
+                                               beam_tables.BASE_TABLE_NAME)
+      if export_gcs:
+        # TODO: define global for bucket
+        destination = beam_tables.get_gcs_folder(dataset, scan_type,
+                                                 'firehook-test')
```
ohnorobo, line 65: We can keep this value for now, but this should be defined in `firehook_resources.py`

pipeline/run_beam_tables.py
```diff
@@ -158,6 +167,11 @@ def parse_args() -> argparse.Namespace:
       default=None,
       metavar='yyyy-mm-dd',
       help='Last date to read data from')
+  parser.add_argument(
+      '--export_gcs',
```
ohnorobo, line 171: We're going to want a way to run these both simultaneously (so we can produce both without having to redo the internal pipeline work.

That's complex to deal with now when the output tables and files are out of sync. But in the future we'll want them to be in sync at all times. Maybe just add a todo for now while we're developing.

ohnorobo (changes requested):

pipeline/run_beam_tables.py
ohnorobo: Also, not a change for this PR. But eventually we're going to want both a prod and dev bucket so we're not writing dev stuff to a prod bucket we probably want to make world-readable.

ohnorobo:
This looks really great. Thanks for navigating the weird complexities of the file writing apis.

pipeline/manual_e2e_test.py
```diff
@@ -270,6 +320,81 @@ def get_bq_rows(client: cloud_bigquery.Client, table_names: List[str]) -> List:
 class PipelineManualE2eTest(unittest.TestCase):
   """Manual tests that require access to cloud project resources."""
 
+  def test_hyperquack_gcs_pipeline_e2e(self) -> None:
```
ohnorobo, line 323: Also add a satellite pipeline test

pipeline/beam_tables.py
avirkud: done

pipeline/beam_tables.py
avirkud: done

pipeline/beam_tables.py
avirkud: done

pipeline/beam_tables.py
avirkud: done

pipeline/run_beam_tables.py
avirkud: Moved to `firehook_resources.py` and added a TODO for separate prod and dev buckets

pipeline/run_beam_tables.py
avirkud: Added a TODO

pipeline/manual_e2e_test.py
avirkud: done

avirkud:
Merged in the schema changes from master. In `schema.py`, renamed `flatten_for_bigquery` to `flatten_to_dict` and added an extra step `dict_to_gcs_json_string` that drops fields we don't want exported to GCS.

pipeline/beam_tables.py
```diff
@@ -354,6 +429,70 @@ def _write_to_bigquery(self, scan_type: str,
...
+        or 'satellite'
+      rows: PCollection[BigqueryRow] of data to write.
+      gcs_folder: GCS folder gs://bucket/folder/... like
+        'gs://firehook-test/scans/echo' to write output files to.
+
+    Raises:
+      Exception: if any arguments are invalid.
+    """
+
+    def _get_destination(record: str) -> str:
```
ohnorobo, line 451: make `_get_destination` and `_custom_file_naming` top-level functions instead of nested here

pipeline/beam_tables.py
```diff
@@ -289,9 +360,13 @@ def _data_to_load(self,
...
     """
     if incremental_load:
-      full_table_name = self._get_full_table_name(table_name)
-      existing_sources = _get_existing_datasources(full_table_name,
-                                                   self.project)
+      if table_name:
+        full_table_name = self._get_full_table_name(table_name)
+        existing_sources = _get_existing_bq_datasources(full_table_name,
+                                                        self.project)
+      elif gcs_folder:
```
ohnorobo, line 367: add an else clause that raises an error if both are None.

ohnorobo (approved):

pipeline/beam_tables.py
avirkud: done

pipeline/beam_tables.py
avirkud: done

avirkud:
Also removed the `outcome` analysis field from the Hyperquack GCS schema.

ohnorobo:
> Also removed the `outcome` analysis field from the Hyperquack GCS schema.

We may eventually want to add this back into the file dumps. We can chat about it in the meeting. I think it's less likely to confuse people than the `anomaly` field, since it's directly based on what we saw happening with the connection (reset, timeout etc.) It's also easier for people to inspect and understand how it's calculated since we have both the [code](https://github.com/censoredplanet/censoredplanet-analysis/blob/master/pipeline/metadata/hyperquack_outcome.py) and [unit tests](https://github.com/censoredplanet/censoredplanet-analysis/blob/master/pipeline/metadata/test_hyperquack_outcome.py) open sourced.

PR #169, opened 2022-06-22
fix outcome calculation fo read emojis

ohnorobo:
Live on the [dashboard](https://datastudio.google.com/c/u/0/reporting/c9afebdd-91c8-4381-9b7d-5a9fd083bb90/page/p_6wx99eilrc)


fortuna (approved):
Thanks for the quick fix!

PR #171, opened 2022-06-28
validate whether cert matches domain in pipeline rather than query

ohnorobo:
This is using the `ssl.match_hostname` method ([docs](https://docs.python.org/3/library/ssl.html#ssl.match_hostname), [code](https://hg.python.org/cpython/file/3.5/Lib/ssl.py#l263)) This has two bad features:
- it's deprecated in favor of just using openssl, but that doesn't expose the exact piece of the interface we need
- it requires mocking out an object similar to the return from `ssl.SSLSocket.getpeercert` since it's really only intended to be used as part of a network call with that function.

But I wanted to avoid having to roll-our-own to avoid handling a bunch of RFC details like unicode domain names.

This is outputting the new fields
- `received_tls_cert_matches_domain` in the hyperquack tables
- `https_tls_cert_matches_domain` in the satellite tables

cc @avirkud I'm currently excluding those fields from the GCS export, but we should eventually add them in since they're straightforward.

(will squash commits down later)


eltsai (approved):

ohnorobo:
Merged as https://github.com/censoredplanet/censoredplanet-analysis/pull/183

PR #173, opened 2022-07-20
pass through trusted cert and matches hostname fields

ohnorobo:
This picks up the
- `trusted_cert`
- `cert_hostname_match`
fields from blockpages.json (new since 2022-06-26) and writes them out to the bq as 
- `https_tls_cert_has_trusted_ca` 
- `https_tls_cert_matches_domain`

This will have some minor conflicts with https://github.com/censoredplanet/censoredplanet-analysis/pull/171 which will need to resolved since they're both working on `https_tls_cert_matches_domain`, but it will be fastest to combine the backfills for both of them together.


ohnorobo:
merged as https://github.com/censoredplanet/censoredplanet-analysis/pull/183

PR #174, opened 2022-07-21
surface cert domain instead of network on bad cert

ohnorobo:
This gives us more specifically useful information about whether the cert is related


table/queries/derived_satellite_scans.sql
```diff
@@ -121,7 +121,7 @@ CREATE TEMP FUNCTION OutcomeString(domain_name STRING,
                       THEN "✅answer:cert_for_domain"
                 WHEN (SELECT LOGICAL_AND(NOT IsCertForDomain(a.https_tls_cert, domain_name))
                       FROM UNNEST(answers) a)
-                      THEN CONCAT("❗️answer:cert_not_for_domain:", AnswersSignature(answers))
+                      THEN CONCAT("❗️answer:cert_not_for_domain:", answers[OFFSET(0)].https_tls_cert_common_name)
```
fortuna, line 124: Perhaps we could say CN=<domain> as that's standard notation. It may be worth listing the first entry in the SAN if the CN is missing.

fortuna (approved):

PR #187, opened 2022-09-14
add ca and cert information to outcomes and change some outcomes to warn

ohnorobo:
- Update cert outcomes to account for the ca validation bit
- Output cert issuer when cert is invalid
- Change some outcome strings to use the ⚠️ emoji instead of ❗

example view [dashboard](https://datastudio.google.com/c/reporting/c9afebdd-91c8-4381-9b7d-5a9fd083bb90): 
image


table/queries/derived_satellite_scans.sql
```diff
@@ -1,52 +1,52 @@
 # https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#dns-parameters-6
 CREATE TEMP FUNCTION ClassifySatelliteRCode(rcode INTEGER) AS (
   CASE
-    WHEN rcode = 1 THEN "❗️dns/rcode:FormErr"
-    WHEN rcode = 2 THEN "❗️dns/rcode:ServFail"
+    WHEN rcode = 1 THEN "⚠️dns/rcode:FormErr"
+    WHEN rcode = 2 THEN "⚠️dns/rcode:ServFail"
     WHEN rcode = 3 THEN "❗️dns/rcode:NXDomain"
```
ramakrishnansr, line 6: Why is the ServFail not as good an indicator of censorship as NXDomain is? Are we worried about geoblocking in the resolution process? Is there a specific example for this?

table/queries/derived_satellite_scans.sql
```diff
@@ -1,52 +1,52 @@
...
-    WHEN REGEXP_CONTAINS(error, '"Err": 111') THEN "❗️read/udp.refused"
-    WHEN REGEXP_CONTAINS(error, '"Err": 113') THEN "❗️read/ip.host_no_route"
-    WHEN REGEXP_CONTAINS(error, '"Err": 24') THEN "❗️setup/system_failure" # Too many open files
-    WHEN error = "{}" THEN "❗️dns/unknown" # TODO figure out origin
-    WHEN error = "no_answer" THEN "❗️dns/answer:no_answer"
+    WHEN REGEXP_CONTAINS(error, '"Err": {}') THEN "⚠️read/udp.timeout"
+    WHEN REGEXP_CONTAINS(error, '"Err": 90') THEN "⚠️read/dns.msgsize"
+    WHEN REGEXP_CONTAINS(error, '"Err": 111') THEN "⚠️read/udp.refused"
+    WHEN REGEXP_CONTAINS(error, '"Err": 113') THEN "⚠️read/ip.host_no_route"
+    WHEN REGEXP_CONTAINS(error, '"Err": 24') THEN "⚠️setup/system_failure" # Too many open files
```
ramakrishnansr, line 38: For these two, wouldn't it be safe to give a different emoji that indicates it could a measurement problems? i.e. in these cases, there is no way one can figure out whether it is censorship according to our measurements. For the others, we are capturing something, but just not sure about what we are capturing. I'm thinking of 4 divisions kind of like OONI's mechanism:
1. Okay
2. Our error
3. Possible censorship but cant be sure
4. Censorship

These two could be assigned code 2.

table/queries/derived_satellite_scans.sql
```diff
@@ -1,52 +1,52 @@
...
-    WHEN ENDS_WITH(error, "read: protocol error") THEN "❗️read/protocol_error"
-    WHEN ENDS_WITH(error, "socket: too many open files") THEN "❗️setup/system_failure"
-    ELSE CONCAT("❗️dns/unknown_error:", error)
+    WHEN ENDS_WITH(error, "i/o timeout") THEN "⚠️read/udp.timeout"
+    WHEN ENDS_WITH(error, "message too long") THEN "⚠️read/dns.msgsize"
+    WHEN ENDS_WITH(error, "connection refused") THEN "⚠️read/udp.refused"
+    WHEN ENDS_WITH(error, "no route to host") THEN "⚠️read/ip.host_no_route"
+    WHEN ENDS_WITH(error, "short read") THEN "⚠️read/dns.msgsize"
+    WHEN ENDS_WITH(error, "read: protocol error") THEN "⚠️read/protocol_error"
+    WHEN ENDS_WITH(error, "socket: too many open files") THEN "⚠️setup/system_failure"
```
ramakrishnansr, line 48: Same as above comment.

ramakrishnansr (approved):
Left some comments that may make messages clearer but overall looks good. Thanks @ohnorobo.

table/queries/derived_satellite_scans.sql
ohnorobo: These aren't set in stone. (Actually I think we're likely to change them a lot as we fine-tune) I'm just basing this on staring at the dashboards a lot for [this doc](https://docs.google.com/document/d/1OnfLgJ3OdghLiPf9tKdqgIl346EehXcZFLrcTtDKwE8/edit#) and noticing that lots of countries have some residue of servfail/refused that doesn't seem to be censorship.

Actually. we can quantify this. Looking at [this view](https://datastudio.google.com/s/v2sohsCeT9g) (just those three Rcode outcomes, fortunately all other codes are much less common)
image
ServFail appears in almost every country, while Refused and Nxdomain appear only in some countries.

Based on that I'm going to switch to
`⚠️dns/rcode:ServFail`
`❗️dns/rcode:NXDomain`
`❗️dns/rcode:Refused`

table/queries/derived_satellite_scans.sql
ohnorobo: If we think something is our error we should just filter it out and not show it in the dashboard.

We actually are filtering out `setup`
https://github.com/censoredplanet/censoredplanet-analysis/pull/187/files#diff-20fce2db9fd5531520462ed659edd5c769af3568b7ff5b03617c44be1434f3d0L156

I've removed the emoji so that will match correctly.

We could remove other things like `ip.host_no_route` as well, but I'd rather do that in a separate change supported by what we're seeing in the data.

PR #193, opened 2022-11-04
add emoji to hyperquack outcomes

ohnorobo:
Side dashboard with the emojified outcomes: [here](https://datastudio.google.com/c/reporting/40dc493c-9974-415c-93ec-bab04de93b6d/page/T33gB)

image

### DIFF
--- a/table/queries/merged_reduced_scans.sql
+++ b/table/queries/merged_reduced_scans.sql
@@ -12,6 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+CREATE TEMP FUNCTION AddOutcomeEmoji(outcome STRING) AS (
+  CASE
+    WHEN STARTS_WITH(outcome, "setup/") THEN outcome
+    WHEN STARTS_WITH(outcome, "expected/") THEN CONCAT("✅", SUBSTR(outcome, 10))
+    ELSE CONCAT("❗️", outcome)
+  END
+);
+
 # BASE_DATASET and DERIVED_DATASET are reserved dataset placeholder names
 # which will be replaced when running the query
 
@@ -49,7 +57,7 @@ WITH AllScans AS (
         server_country AS country_code,
         server_as_full_name AS network,
         IF(domain_is_control, "CONTROL", domain) AS domain,
-        outcome AS outcome,
+        AddOutcomeEmoji(outcome) AS outcome,
         CONCAT("AS", server_asn, IF(server_organization IS NOT NULL, CONCAT(" - ", server_organization), "")) AS subnetwork,
         IFNULL(domain_category, "Uncategorized") AS category,
         COUNT(*) AS count
@@ -64,8 +72,8 @@ SELECT
     Grouped.* EXCEPT (country_code),
     IFNULL(country_name, country_code) AS country_name,
     CASE
-        WHEN (STARTS_WITH(outcome, "expected/")) THEN 0
-        WHEN (STARTS_WITH(outcome, "dial/") OR STARTS_WITH(outcome, "setup/") OR ENDS_WITH(outcome, "/invalid")) THEN NULL
+        WHEN (STARTS_WITH(outcome, "✅")) THEN 0
+        WHEN (STARTS_WITH(outcome, "❗️dial/") OR STARTS_WITH(outcome, "setup/") OR ENDS_WITH(outcome, "/invalid")) THEN NULL
         WHEN (ENDS_WITH(outcome, "unknown")) THEN count / 2.0
         ELSE count
     END AS unexpected_count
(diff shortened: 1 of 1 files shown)


table/queries/merged_reduced_scans.sql
```diff
@@ -12,6 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+CREATE TEMP FUNCTION AddOutcomeEmoji(outcome STRING) AS (
+  CASE
+    WHEN STARTS_WITH(outcome, "setup/") THEN outcome
```
fortuna, line 17: We were using ❔ for invalid tests. I think we should have prefixes for all of them. This was it's easier to say that those outcomes are ignored for the expected rate.

table/queries/merged_reduced_scans.sql
```diff
@@ -64,8 +72,8 @@ SELECT
     Grouped.* EXCEPT (country_code),
     IFNULL(country_name, country_code) AS country_name,
     CASE
-        WHEN (STARTS_WITH(outcome, "expected/")) THEN 0
-        WHEN (STARTS_WITH(outcome, "dial/") OR STARTS_WITH(outcome, "setup/") OR ENDS_WITH(outcome, "/invalid")) THEN NULL
+        WHEN (STARTS_WITH(outcome, "✅")) THEN 0
+        WHEN (STARTS_WITH(outcome, "❗️dial/") OR STARTS_WITH(outcome, "setup/") OR ENDS_WITH(outcome, "/invalid")) THEN NULL
         WHEN (ENDS_WITH(outcome, "unknown")) THEN count / 2.0
         ELSE count
```
fortuna, line 78: Let's use ❗️ for unexpected results only, so it's clearer. It would be nice to see something like this, with outcome aligned with expected evaluation:

```suggestion
        WHEN (STARTS_WITH(outcome, "✅")) THEN 0
        WHEN (STARTS_WITH(outcome, "❗️")) THEN count
        WHEN (STARTS_WITH(outcome, "❔")) THEN NULL
        ELSE count / 2.0  # unknown
```

fortuna (changes requested):

table/queries/merged_reduced_scans.sql
ohnorobo: added prefixes for all

table/queries/merged_reduced_scans.sql
```diff
@@ -58,22 +67,25 @@ WITH AllScans AS (
...
     Grouped.* EXCEPT (country_code),
     IFNULL(country_name, country_code) AS country_name,
     CASE
-        WHEN (STARTS_WITH(outcome, "expected/")) THEN 0
-        WHEN (STARTS_WITH(outcome, "dial/") OR STARTS_WITH(outcome, "setup/") OR ENDS_WITH(outcome, "/invalid")) THEN NULL
-        WHEN (ENDS_WITH(outcome, "unknown")) THEN count / 2.0
-        ELSE count
+        WHEN (STARTS_WITH(outcome, "✅")) THEN 0
+        WHEN (STARTS_WITH(outcome, "❗️")) THEN count
+        WHEN (STARTS_WITH(outcome, "❔")) THEN NULL
```
ohnorobo, line 78: the only remaining ❔ outcome that is in the final table is now `❔unknown`.

I've switched that to NULL instead of count/2 on the grounds that count/2 never made any sense. unknown is by definition an error in our data handling.

fortuna: unknown was for errors we don't know how to process.
I worry that we will be ignoring things we shouldn't.
The goal of making them 1/2 was to draw attention to them. Because anything that is not ~0 or ~100% needs further investigation, so using 1/2 pulls the results to more investigation, rather than hiding them one way or another.
Perhaps we need to take a look if we have any more unknowns in the data and where they are coming from.

fortuna: Perhaps you can use ± or 🤷 for unknown.

ohnorobo: Currently unknown is the only outcome type that will have the ❔ emoji, since setup is already entirely filtered out of the dashboard. That's what will help us notice it. Having fractions in the unexpected counts will not help because in practice many of them are already fractional.

There are currently only 2 measurements marked as unknown in the entire hyperquack dataset going back to 2018. [data](https://docs.google.com/spreadsheets/d/1hLTH83D-7QQXgpE0hLNi4E7_S5GaCUGllsCbfZq63-M/edit#gid=1809681870)

❔ is a reasonable way to differentiate this case. The other symbols seem overly silly.

table/queries/merged_reduced_scans.sql
```diff
@@ -58,22 +67,25 @@ WITH AllScans AS (
     WHERE NOT controls_failed
     GROUP BY date, source, country_code, network, outcome, domain, category, subnetwork
     # Filter it here so that we don't need to load the outcome to apply the report filtering on every filter.
-    HAVING NOT STARTS_WITH(outcome, "setup/")
+    HAVING NOT STARTS_WITH(outcome, "❔setup/")
```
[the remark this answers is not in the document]
fortuna: Shouldn't we filter out all the NULL results then?
```suggestion
    HAVING NOT STARTS_WITH(outcome, "❔")
```

fortuna (approved):

PR #199, opened 2022-11-18
Retries

ohnorobo:
This adds a `retry` field to the satellite and hyperquack datasets. This represents the retry index for that measurement from 0 to N. Control measurements have a retry index of `None`.

We can use this later to filter out any retries measurements except the final one.

For some of the weirder old satellite formats this gets a bit complex. We're actually [not even reading](https://github.com/censoredplanet/censoredplanet-analysis/blob/ac471e9ca9bd2ef1466ad0a326dbff0ec0498e13/pipeline/beam_tables.py#L66) the data from this period, but I've includes the parsing here for correctness.

Ran a [test backfill](https://console.cloud.google.com/dataflow/jobs/us-central1/2022-11-22_08_28_27-3382697350724504316?project=firehook-censoredplanet)

Here are the [enumerated retry outcome sequences](https://docs.google.com/spreadsheets/d/1jOIY_xxp51xFw5lrbbwicgThentYrMPtZMDCe71gX28/edit#gid=702521458) we see in the output data.


pipeline/metadata/flatten_hyperquack.py
```diff
@@ -186,3 +192,5 @@ def _process_hyperquack_v2(self, filename: str, scan: Any,
         row.error = response['error']
 
       yield row
+      if not is_control:
+        retry_index += 1
```
ramakrishnansr, line 196: Is this not better to put before the yield?

ramakrishnansr (approved):

pipeline/metadata/flatten_hyperquack.py
ohnorobo: No, this would cause an off-by-one error. The retry indexes start at 0.

PR #195, opened 2022-11-09
filter bad resolvers

ohnorobo:
Roughly filter DNS resolvers that aren't reliable because they have too many errors.

Dashboards: [1](https://datastudio.google.com/c/u/0/reporting/bcba5de4-0402-4687-8b7c-1ad1b04ad3d1/page/p_0foeibtt0c), [2](https://datastudio.google.com/c/reporting/fa204158-932b-43bf-b3e8-b346c1468792/page/p_mekso35lrc)

This is better than the [previous filtering solution](https://github.com/censoredplanet/censoredplanet-analysis/commit/4a28e74bf2aa13cdacba48523de7500c46207661) which was just a hacky query.

PR #200, opened 2022-11-24
retry filtering query

ohnorobo:
Filter out all rountrips except the one with the highest retry in the measurement. This means for a measurement like
`timeout`, `timeout`, `success`
we will only show the final success
and for `timeout`, `timeout`, `timeout`, `timeout`
we will only show the single final timeout.
Domain control measurement have a retry index of `NULL` but we still include them.

This cuts the total timeouts in the data approximately 2/3, from ~6% to ~2%. 
(We already got a big improvement in timeout number from removing bad resolvers in https://github.com/censoredplanet/censoredplanet-analysis/pull/195)

With only final retries, [dashboard](https://datastudio.google.com/c/u/0/reporting/fa204158-932b-43bf-b3e8-b346c1468792/page/p_mekso35lrc): (view from Nov 9th to Dec 1st)
image

Before [dashboard](https://datastudio.google.com/c/reporting/c9afebdd-91c8-4381-9b7d-5a9fd083bb90/page/p_mekso35lrc):
image

This is a follow up to https://github.com/censoredplanet/censoredplanet-analysis/pull/199, which adds the `retry` field.

Something about this join still seems to be dropping some successful measurements despite the distribution of retry outcomes being [very clean](https://docs.google.com/spreadsheets/d/1jOIY_xxp51xFw5lrbbwicgThentYrMPtZMDCe71gX28/edit#gid=702521458).
See: `matches_ip` going from 131mil to 123mil. But it's not the ~50% drop we were seeing before.

### DIFF
--- a/table/queries/derived_satellite_scans.sql
+++ b/table/queries/derived_satellite_scans.sql
@@ -133,6 +133,23 @@ CREATE TEMP FUNCTION OutcomeString(domain_name STRING,
 );
 
 
+# When some roundtrips in the DNS scans fail they retry.
+# These retried are identified in bigquery by having the same measurement id
+# Once a roundtrip succeeds we stop retrying
+# To avoid overcounting retries (ex: measuring x4 timeouts on failure but only x1 success when succeeding)
+# We only want to include the last retry in any set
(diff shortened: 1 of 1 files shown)


fortuna (approved):
I think it would be easier if the number of requests was output in the flattening stage, but this works too.

PR #206, opened 2022-12-05
preliminary emoji changes

ohnorobo:
Tentative proposal for which emoji to use in the outcome strings.

The new set of emojis (in order of severity) is:
- :white_check_mark: expected
- :grey_question: invalid test
- :question: unexpected
- :exclamation: unexpected and suspicious

Changes:
- :warning: -> :question: as being less confusing
- - This means the two unexpected outcomes are :question: and :exclamation: which are pleasingly visually similar
- dial outcomes in hyperquack are now :grey_question: Am I missing other outcomes that should change to this?
- @ramakrishnansr I think you wanted to downgrade more hyperquack outcomes to :question:. [Which ones](https://github.com/censoredplanet/censoredplanet-analysis/blob/master/pipeline/metadata/test_hyperquack_outcome.py)?

Overall Satellite outcomes ([source](https://datastudio.google.com/c/reporting/c9afebdd-91c8-4381-9b7d-5a9fd083bb90/page/p_mekso35lrc)):
image

Overall Hyperquack outcomes ([source](https://datastudio.google.com/c/reporting/f85a1d2f-23f3-4ccf-876e-8d1338580559/page/p_intendmtxc)):
image

### DIFF
--- a/table/queries/derived_satellite_scans.sql
+++ b/table/queries/derived_satellite_scans.sql
@@ -1,30 +1,30 @@
 # https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#dns-parameters-6
 CREATE TEMP FUNCTION ClassifySatelliteRCode(rcode INTEGER) AS (
   CASE
-    WHEN rcode = 1 THEN "❓dns/rcode:FormErr"
+    WHEN rcode = 1 THEN "❗️dns/rcode:FormErr"
     WHEN rcode = 2 THEN "❓dns/rcode:ServFail"
     WHEN rcode = 3 THEN "❗️dns/rcode:NXDomain"
-    WHEN rcode = 4 THEN "❓dns/rcode:NotImp"
+    WHEN rcode = 4 THEN "❗️dns/rcode:NotImp"
     WHEN rcode = 5 THEN "❗️dns/rcode:Refused"
-    WHEN rcode = 6 THEN "❓dns/rcode:YXDomain"
-    WHEN rcode = 7 THEN "❓dns/rcode:YXRRSet"
-    WHEN rcode = 8 THEN "❓dns/rcode:NXRRSet"
-    WHEN rcode = 9 THEN "❓dns/rcode:NotAuth"
-    WHEN rcode = 10 THEN "❓dns/rcode:NotZone"
-    WHEN rcode = 11 THEN "❓dns/rcode:DSOTYPENI"
-    WHEN rcode = 12 THEN "❓dns/rcode:Unassigned"
-    WHEN rcode = 13 THEN "❓dns/rcode:Unassigned"
-    WHEN rcode = 14 THEN "❓dns/rcode:Unassigned"
-    WHEN rcode = 15 THEN "❓dns/rcode:Unassigned"
-    WHEN rcode = 16 THEN "❓dns/rcode:BadVers"
-    WHEN rcode = 17 THEN "❓dns/rcode:BadSig"
-    WHEN rcode = 18 THEN "❓dns/rcode:BadKey"
-    WHEN rcode = 19 THEN "❓dns/rcode:BadTime"
-    WHEN rcode = 20 THEN "❓dns/rcode:BadMode"
-    WHEN rcode = 21 THEN "❓dns/rcode:BadAlg"
-    WHEN rcode = 22 THEN "❓dns/rcode:BadTrunc"
-    WHEN rcode = 23 THEN "❓dns/rcode:BadCookie"
-    ELSE CONCAT("❓dns/unknown_rcode:", rcode)
+    WHEN rcode = 6 THEN "❗️dns/rcode:YXDomain"
+    WHEN rcode = 7 THEN "❗️dns/rcode:YXRRSet"
+    WHEN rcode = 8 THEN "❗️dns/rcode:NXRRSet"
+    WHEN rcode = 9 THEN "❗️dns/rcode:NotAuth"
+    WHEN rcode = 10 THEN "❗️dns/rcode:NotZone"
+    WHEN rcode = 11 THEN "❗️dns/rcode:DSOTYPENI"
+    WHEN rcode = 12 THEN "❗️dns/rcode:Unassigned"
+    WHEN rcode = 13 THEN "❗️dns/rcode:Unassigned"
+    WHEN rcode = 14 THEN "❗️dns/rcode:Unassigned"
+    WHEN rcode = 15 THEN "❗️dns/rcode:Unassigned"
+    WHEN rcode = 16 THEN "❗️dns/rcode:BadVers"
+    WHEN rcode = 17 THEN "❗️dns/rcode:BadSig"
+    WHEN rcode = 18 THEN "❗️dns/rcode:BadKey"
+    WHEN rcode = 19 THEN "❗️dns/rcode:BadTime"
+    WHEN rcode = 20 THEN "❗️dns/rcode:BadMode"
+    WHEN rcode = 21 THEN "❗️dns/rcode:BadAlg"
+    WHEN rcode = 22 THEN "❗️dns/rcode:BadTrunc"
+    WHEN rcode = 23 THEN "❗️dns/rcode:BadCookie"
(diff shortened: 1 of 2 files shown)


fortuna (approved):

fortuna:
Perhaps we can call them:
✅ expected
❔ invalid test
❓ unexpected, but inconclusive
❗ unexpected

❓ indicates methodology gaps. Outcomes for which we found issues (we should document them in the sql files). Hopefully we can address most of the issues and turn ❓  into ❗. For example, if we change the scans to only output ServFails that are repeated, we may be able to change it to ❗.  It seems timeouts have improved a lot, we should check if it still returns false positives. And we need to figure out a way to address the CDN blocking external requests.

table/queries/derived_satellite_scans.sql
```diff
@@ -1,52 +1,52 @@
...
+    WHEN rcode = 6 THEN "❗️dns/rcode:YXDomain"
+    WHEN rcode = 7 THEN "❗️dns/rcode:YXRRSet"
+    WHEN rcode = 8 THEN "❗️dns/rcode:NXRRSet"
+    WHEN rcode = 9 THEN "❗️dns/rcode:NotAuth"
+    WHEN rcode = 10 THEN "❗️dns/rcode:NotZone"
+    WHEN rcode = 11 THEN "❗️dns/rcode:DSOTYPENI"
+    WHEN rcode = 12 THEN "❗️dns/rcode:Unassigned"
+    WHEN rcode = 13 THEN "❗️dns/rcode:Unassigned"
+    WHEN rcode = 14 THEN "❗️dns/rcode:Unassigned"
+    WHEN rcode = 15 THEN "❗️dns/rcode:Unassigned"
```
ramakrishnansr, line 18: Are we sure that these are all cases of censorship?

table/queries/derived_satellite_scans.sql
```diff
@@ -1,52 +1,52 @@
...
-    WHEN REGEXP_CONTAINS(error, '"Err": {}') THEN "⚠️read/udp.timeout"
-    WHEN REGEXP_CONTAINS(error, '"Err": 90') THEN "⚠️read/dns.msgsize"
-    WHEN REGEXP_CONTAINS(error, '"Err": 111') THEN "⚠️read/udp.refused"
-    WHEN REGEXP_CONTAINS(error, '"Err": 113') THEN "⚠️read/ip.host_no_route"
-    WHEN REGEXP_CONTAINS(error, '"Err": 24') THEN "setup/system_failure" # Too many open files
-    WHEN error = "{}" THEN "⚠️dns/unknown" # TODO figure out origin
-    WHEN error = "no_answer" THEN "⚠️dns/answer:no_answer"
+    WHEN REGEXP_CONTAINS(error, '"Err": {}') THEN "❓read/udp.timeout"
+    WHEN REGEXP_CONTAINS(error, '"Err": 90') THEN "❗️read/dns.msgsize"
+    WHEN REGEXP_CONTAINS(error, '"Err": 111') THEN "❗️read/udp.refused"
```
ramakrishnansr, line 36: See comment on refused above

table/queries/merged_reduced_scans.sql
```diff
@@ -67,14 +71,16 @@ WITH AllScans AS (
...
+    HAVING (NOT STARTS_WITH(outcome, "❔setup/")
+            AND NOT outcome = "❔read/system")
 )
 SELECT
     Grouped.* EXCEPT (country_code),
     IFNULL(country_name, country_code) AS country_name,
     CASE
         WHEN (STARTS_WITH(outcome, "✅")) THEN 0
         WHEN (STARTS_WITH(outcome, "❗️")) THEN count
+        WHEN (STARTS_WITH(outcome, "❓")) THEN count
```
ramakrishnansr, line 83: What is the reasoning behind counting these cases? Ideally, we only want to count cases that we're more sure about.

ramakrishnansr:
Ideally, these are the general categories I would put under suspicious for hyperquack:

1. read/write tcp.reset
2. read tcp.timeout
3. content/blockpage:
4. read http.empty

All the dial and setup failures should be marked as measurement failures. Everything else should get unexpected.

table/queries/derived_satellite_scans.sql
```diff
@@ -1,52 +1,52 @@
...
 CREATE TEMP FUNCTION ClassifySatelliteRCode(rcode INTEGER) AS (
   CASE
-    WHEN rcode = 1 THEN "⚠️dns/rcode:FormErr"
-    WHEN rcode = 2 THEN "⚠️dns/rcode:ServFail"
+    WHEN rcode = 1 THEN "❗️dns/rcode:FormErr"
+    WHEN rcode = 2 THEN "❓dns/rcode:ServFail"
     WHEN rcode = 3 THEN "❗️dns/rcode:NXDomain"
-    WHEN rcode = 4 THEN "⚠️dns/rcode:NotImp"
+    WHEN rcode = 4 THEN "❗️dns/rcode:NotImp"
     WHEN rcode = 5 THEN "❗️dns/rcode:Refused"
```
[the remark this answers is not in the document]
fortuna: Refused is unexpected nevertheless for a functioning site.

We must have a clear criteria to decide on the icons. What we landed on was to always use ❗️ for unexpected results, unless either of these are true:

1. There's an expectation of a certain level of failures, like timeouts of Servfail
2. We are not completely sure about what to expect, as in page body match

Those indicate gaps in the methodology, and merit the dubious expectation marker.
We should be documenting all those cases though.

We shouldn't use the dubious marker unless we have documented cases where the expectation is not clear.

We can discuss tweaking the criteria, but it must be clear.

table/queries/derived_satellite_scans.sql
fortuna: We are not, but we also don't have evidence that this is problematic. So it takes the ❗️ marker as per the criteria I mentioned.

table/queries/merged_reduced_scans.sql
fortuna: We still consider those results as unexpected, so we count them as unexpected.
It's just that the expectation is not super clear with those.

table/queries/derived_satellite_scans.sql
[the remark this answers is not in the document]
fortuna: By the way, with the Servfail retry logic @ramakrishnansr added, we should consider using ❗️ for it once it lands in production, since it's clearly unexpected to have it fail multiple times in a row.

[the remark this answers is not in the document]
ramakrishnansr: I don't think Refused is a clear sign of failure for a working site - As mentioned in the link shared above, and based on manual investigation of the data, the appearance of the Refused error seems to be non-deterministic, and hence should warrant the dubious marker. 

The criteria can be: 
1. Deterministic failure for a certain domain where it is inaccessible - Unexpected
2. Non-deterministic failure where the failure may not be due to the domain - Dubious.

table/queries/merged_reduced_scans.sql
ramakrishnansr: But maybe they lead to certain dubious cases bubbling up, when it may actually be worthwhile to bubble up ones we're more confident about. What about creating a filter where the count can be generated dynamically based on the filter?

ohnorobo: Counting these differently would be a very large change to the way we're displaying the data in the pivot table. I'm not fully opposed to it. But I wouldn't want to do it as part of this change. It would have to be its own thing with its own separate analysis.

ohnorobo:
> Ideally, these are the general categories I would put under suspicious for hyperquack:

Sounds good. I've implemented that like [this](https://github.com/censoredplanet/censoredplanet-analysis/pull/206/commits/eaec8168b3a49b1fa001a7369183887a9058575c).

Resulting in these outcomes:
image

The main change is that the `tls/` and `http/` outcomes are now ❓.

`read/http_truncated_response` is still ❗. But this is such a small outcome I think it's worth it for the simpler case logic it allows.

ramakrishnansr:
LGTM. Thanks.

PR #219, opened 2023-02-21
Nightly GCS export

agiix:


pipeline/beam_tables.py
```diff
@@ -542,7 +553,8 @@ def run_beam_pipeline(self, scan_type: str, incremental_load: bool,
                         gcs_folder: Optional[str],
                         start_date: Optional[datetime.date],
                         end_date: Optional[datetime.date],
-                        export_gcs: bool) -> None:
+                        export_gcs: bool,
+                        bq_and_gcs: bool) -> None:
```
ohnorobo, line 557: I think out would be best to have 

export_gcs and export_bq 

as separate bool options, and then having both false raise an error.

ohnorobo (changes requested):
advice for passing commit hooks: https://github.com/censoredplanet/censoredplanet-analysis/blob/master/docs/development.md

pipeline/beam_tables.py
```diff
@@ -597,7 +611,10 @@ def run_beam_pipeline(self, scan_type: str, incremental_load: bool,
 
       # TODO: Eventually we want to be able to run both exports simultaneously
       # so that we can reuse the pipeline output.
-      if export_gcs and gcs_folder is not None:
+      if bq_and_gcs:
+        self._write_to_bigquery(scan_type, rows, table_name, incremental_load)
```
ohnorobo, line 615: I would just leave this to-be-implemented for now and raise an exception. We can't test it until we have the outputs in sync.

pipeline/beam_tables.py
```diff
@@ -597,7 +611,10 @@ def run_beam_pipeline(self, scan_type: str, incremental_load: bool,
 
       # TODO: Eventually we want to be able to run both exports simultaneously
```
ohnorobo, line 612: delete this todo

pipeline/beam_tables.py
agiix: Done

pipeline/beam_tables.py
agiix: I implemented a simple function for that, but I can remove it again for now.

pipeline/beam_tables.py
agiix: Done

pipeline/beam_tables.py
```diff
@@ -509,6 +536,57 @@ def _write_to_gcs(self, scan_type: str,
         max_writers_per_bundle=0,
         file_naming=_custom_file_naming(suffix='.json.gz')))
 
+  def _write_to_bq_and_gcs(self, scan_type: str,
```
ohnorobo, line 539: maybe refactor out the common pieces for this into write_to_bq and _write_to_gcs individual functions. Or just call both, since calling them serially doesn't stop them from running in parallel on execution.

pipeline/run_beam_tables.py
```diff
@@ -199,7 +208,12 @@ def parse_args() -> argparse.Namespace:
       '--export_gcs',
       action='store_true',
       default=False,
-      help='Export to Google Cloud Storage instead of BigQuery')
+      help='Export to Google Cloud Storage')
+  parser.add_argument(
+      '--export_bq',
```
ohnorobo, line 213: add this argument to schedule.py

pipeline/beam_tables.py
```diff
@@ -397,7 +415,17 @@ def _data_to_load(self,
...
     """
     if incremental_load:
-      if table_name:
+      if table_name and gcs_folder:
+        full_table_name = self._get_full_table_name(table_name)
+        table_existing_sources = _get_existing_bq_datasources(
+            full_table_name, self.project)
+        gcs_existing_sources = _get_existing_gcs_datasources(
+            gcs_folder, self.project)
+        table_set = set(table_existing_sources)
```
ohnorobo, line 424: nit: put these set operations inside of compare_sets and make it just compare_lists.

pipeline/beam_tables.py
```diff
@@ -372,6 +374,22 @@ def _get_full_table_name(self, table_name: str) -> str:
...
+    if table_set != gcs_set:
+      table_diff = table_set.difference(gcs_set)
+      diff_exception = ''
+      if table_diff:
+        diff_exception += '\nThe following files are missing in BigQuery:\n'
+        missing_files = ('\n').join(sorted(table_diff))
+        diff_exception += missing_files
+      gcs_diff = gcs_set.difference(table_set)
+      if gcs_diff:
+        diff_exception += '\nThe following files are missing in the GCS folder:\n'
```
ohnorobo, line 387: Assuming that this currently fails when running a pipeline, can you add a comment/pipeline link in the PR description explaining the current failure? (No need to fix the files)

pipeline/beam_tables.py
```diff
@@ -372,6 +374,22 @@ def _get_full_table_name(self, table_name: str) -> str:
     """
     return self.project + ':' + table_name
 
+  def _compare_sets(self, table_set: set, gcs_set: set) -> None:
```
ohnorobo, line 377: add a docstring (see elsewhere in this file) explaining the function, types and especially the raised errors

pipeline/beam_tables.py
```diff
@@ -397,7 +415,17 @@ def _data_to_load(self,
...
-      if table_name:
+      if table_name and gcs_folder:
+        full_table_name = self._get_full_table_name(table_name)
+        table_existing_sources = _get_existing_bq_datasources(
+            full_table_name, self.project)
+        gcs_existing_sources = _get_existing_gcs_datasources(
+            gcs_folder, self.project)
+        table_set = set(table_existing_sources)
+        gcs_set = set(gcs_existing_sources)
+        self._compare_sets(table_set, gcs_set)
```
ohnorobo, line 426: only check this when planning to write to both datasources

pipeline/beam_tables.py
agiix: I agree, lets just keep the separate functions and call them each.

pipeline/run_beam_tables.py
agiix: Done.

pipeline/beam_tables.py
agiix: Done.

pipeline/beam_tables.py
agiix: It does already, since it only enters this branch when table_name and gcs_folder have been set.

ohnorobo (approved):
approved, still need to remove the schedule.py arg

ohnorobo (approved):

PR #233, opened 2023-05-05
Filter non-https domains

ohnorobo:
Remove any domains which don't have at least one HTTPS cert match/valid-ca in a given scan.

We don't have to know exactly why any given domain is failing (long dead, bad cert that day, fat-finger DNS, never supported HTTPS, etc) but if we can't confirm one good cert than we have to remove it.

This is possible now because we are also getting the page fetches for control answers [here](https://github.com/censoredplanet/censoredplanet-analysis/pull/231).

### DIFF
--- a/table/queries/derived_satellite_scans.sql
+++ b/table/queries/derived_satellite_scans.sql
@@ -154,10 +154,10 @@ AS (
 
 
 # Get all domains that have even a single valid HTTPS certificate resolution per scan
-CREATE OR REPLACE TABLE `firehook-censoredplanet.DERIVED_DATASET.https_capable_domains`
+CREATE OR REPLACE TABLE `PROJECT_NAME.DERIVED_DATASET.https_capable_domains`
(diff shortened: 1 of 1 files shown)


table/queries/derived_satellite_scans.sql
```diff
@@ -153,6 +153,17 @@ AS (
 );
 
 
+# Get all domains that have even a single valid HTTPS certificate resolution per scan
+CREATE OR REPLACE TABLE `PROJECT_NAME.DERIVED_DATASET.https_capable_domains`
```
fortuna, line 157: Add a TODO to do this in the pipeline, so it's cheaper and incremental.

fortuna (approved):

table/queries/derived_satellite_scans.sql
ohnorobo: done

PR #235, opened 2023-05-10
Split validation

ohnorobo:
This splits `answer:not_validated` into `answer:unvalidated_http_connection` and `answer:no_tcp_connection` (when we were not able to connect to the answer ip at all.

I'd like suggestions about:
- the clearest (short) names to use for these outcome
- thoughts on this approach overall. I think trying to break out `not_validated` further is useful, since it's 60% of our unexpected outcomes. But I don't have a full theory of how this new information is actionable.
- Do we want to break out `answer:no_tcp_connection` further, by splitting along different types of connection errors? (timeout, no route, etc)

dashboard with this data: [here](https://lookerstudio.google.com/c/u/0/reporting/d6e93931-f621-46cb-8f37-1a8186db7be4/page/p_frq2r0hs5c?params=%7B%22df33%22:%22include%25EE%2580%25800%25EE%2580%2580IN%25EE%2580%2580%25E2%259D%2597%25EF%25B8%258Fpage%253Ahttp_blockpage%22%7D)

Here's the new outcome breakdown
image

### DIFF
--- a/table/queries/derived_satellite_scans.sql
+++ b/table/queries/derived_satellite_scans.sql
@@ -28,6 +28,66 @@ CREATE TEMP FUNCTION ClassifySatelliteRCode(rcode INTEGER) AS (
   END
 );
 
+# Classify all page fetch errors into a small set of enums
+#
+# Output is a string of the format "stage/outcome"
+# Documentation of this enum is at
+# https://github.com/censoredplanet/censoredplanet-analysis/blob/master/docs/tables.md#outcome-classification
+CREATE TEMP FUNCTION ClassifyPageFetchError(error STRING) AS (
+  CASE
+    # System failures
+    WHEN ENDS_WITH(error, "address already in use") THEN "setup/system_failure"
+    WHEN ENDS_WITH(error, "protocol error") THEN "setup/system_failure"
+    WHEN ENDS_WITH(error, "protocol not available") THEN "setup/system_failure" # ipv4 vs 6 error
+    WHEN ENDS_WITH(error, "too many open files") THEN "setup/system_failure"
+
+    # Dial failures
+    WHEN ENDS_WITH(error, "network is unreachable") THEN "dial/ip.network_unreachable"
+    WHEN ENDS_WITH(error, "no route to host") THEN "dial/ip.host_no_route"
+    WHEN ENDS_WITH(error, "connection refused") THEN "dial/tcp.refused"
+    WHEN ENDS_WITH(error, "context deadline exceeded") THEN "dial/timeout"
+    WHEN ENDS_WITH(error, "connect: connection timed out") THEN "dial/timeout"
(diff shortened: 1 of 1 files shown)


table/queries/derived_satellite_scans.sql
```diff
@@ -129,7 +129,15 @@ CREATE TEMP FUNCTION OutcomeString(domain_name STRING,
                 WHEN (SELECT LOGICAL_OR(answer.matches_control.asn)
                       FROM UNNEST(answers) answer)
                       THEN "✅answer:matches_asn"
-                ELSE CONCAT("❓answer:not_validated:", AnswersSignature(answers))
+                # TODO: delete, this case is covered by all the cert cases above
+                WHEN (SELECT LOGICAL_OR(NOT (a.https_response_status IS NULL))
```
fortuna, line 133: Can we tell whether we are able to establish a TCP connection to port 443?

I would suggest to create a "{error}:{AS}", and move it to right after `matches:ip`. Possible errors: connection_timeout, address_unreachable, network_unreachable, connection_refused.
Example: `connection_timeout:TWITTER` or maybe `answer:TWITTER:connection_refused`

We would still keep the not_validated as an unexpected catch all in that case.

fortuna: BTW, you can find the syscall error strings from Go at https://cs.opensource.google/go/go/+/master:src/syscall/zerrors_linux_amd64.go

fortuna (changes requested):

table/queries/derived_satellite_scans.sql
```diff
@@ -129,7 +129,15 @@ CREATE TEMP FUNCTION OutcomeString(domain_name STRING,
...
+                      FROM UNNEST(answers) a)
+                      THEN CONCAT("❗️answer:unvalidated_https_connection:", AnswersSignature(answers))
+                WHEN (SELECT LOGICAL_OR(NOT (a.http_response_status IS NULL))
```
fortuna, line 136: I also suggest we drop the HTTP test. It makes it harder to reason about the methodology by involving another port and another fetch.

fortuna (changes requested):

ohnorobo:
So @fortuna the overall thought here is:
- drop HTTP page fetches. We know the domain is capable of HTTPS, so what happens with port 80 on a random dubiously-received IP is not helpful.
- this will also drop HTTP blockpages, which by now are very small fraction of outcomes. (0.14% of unexpected)
- Any IP where we received a response on port 443 is split out into one of the cert in/valid outcomes.
- Any IP where we did not get a response on 443 (currently marked as `not_validated`) gets split out by error type (timeout, unreachable, refused, etc)

ohnorobo:
I made a sample dashboard that does something like this [here](https://lookerstudio.google.com/c/u/0/reporting/6327bd58-8696-4937-89bc-9c7a55c80efe/page/p_mekso35lrc)

I'm not yet convinced that having the errors allows you to make any better analysis decisions, but it's interesting to be able to see.

image

PR #241, opened 2023-06-06
bump up domain clustering

ohnorobo:
Cluster earlier by domain. This will make FE queries generally cheaper since we have domain filtering in the default view.


ohnorobo:
merging so we have everything in place to turn nightly back on

table/queries/merged_reduced_scans.sql
```diff
@@ -38,7 +38,7 @@ CREATE OR REPLACE TABLE `PROJECT_NAME.DERIVED_DATASET.merged_reduced_scans_v2`
 PARTITION BY date
 # Columns `source` and `country_name` are always used for filtering and must come first.
 # `network` and `domain` are useful for filtering and grouping.
```
fortuna, line 40: It may be worth mentioning that the order matters and link to https://cloud.google.com/bigquery/docs/clustered-tables#cluster_column_ordering
Also that we need domain to come after country to benefit from the default domain selection.

PR #242, opened 2023-06-06
run weekly instead of daily

ohnorobo:
Another temporary measure for turning nightly on (merging to `prod-side`). This will run the pipeline weekly instead of daily.


ramakrishnansr:
Okay temporarily. Is this Monday 4:00 AM ET? Our Satellite runs finish around 12:30 AM Sundays and I want to make sure we capture that.

ohnorobo:
I'm also suggesting we don't run the satellite query in prod in https://github.com/censoredplanet/censoredplanet-analysis/pull/243. It will still run in dev.

PR #251, opened 2023-08-25
remove reshuffle

ohnorobo:
This change is the opposite of https://github.com/censoredplanet/censoredplanet-analysis/pull/224 since reshuffling has been causing issues with workers loosing contact in prod.


pipeline/beam_tables.py
```diff
@@ -622,12 +622,6 @@ def run_beam_pipeline(self, scan_type: str, incremental_load: bool,
       # PCollection[Tuple[filename,line]]
       lines = _read_scan_text(p, new_filenames)
 
-      # Reshuffle to break fusion on long files
```
fortuna, line 625: Do you wanna comment it out instead and leave a comment?

fortuna (approved):

pipeline/beam_tables.py
ohnorobo: I think the better place to record it as a possibility is in the pull requests adding/removing it, not as a lingering code comment.

